PR #89, opened 2016-07-07
enable multiple connections to same host via distinct app_name

jjallaire:


tarunparmar:
Is this the issue where if a connection is already established from a specific user and app_*  then another user+app_* cannot connect and will show as WAITING on spark UI

The WAITING app will connect once the RUNNING app is disconnected.

image

PR #106, opened 2016-07-13
support for reading from relative paths

jjallaire:
Call normalizePath rather than path.expand for reading/writing data (this gives us relative path resolution as well as "~" resolution). This is principally useful for local mode as within a cluster fully qualfied paths to the filesystem, S3, or HDFS are used.

@tareefk


javierluraschi:
LKTM

PR #138, opened 2016-07-29
extend spark_config documentation with a link to config-template.yml

MarcinKosinski:
Not every user knows where is that default specification :)


jjallaire:
Note that you don't need the `file.path` call (`system.file` will also append together a variable list of path segments). What about just linking to the source file on GitHub: https://github.com/rstudio/sparklyr/blob/master/inst/conf/config-template.yml

MarcinKosinski:
moved to https://github.com/rstudio/sparklyr/pull/141

Issue #136, opened 2016-07-29
defaul config YAML for cluster deployment


MarcinKosinski:
So far you have default `config-template.yml` for local Spark applications https://github.com/rstudio/sparklyr/blob/master/inst/conf/config-template.yml

According to #115 and #131, have you thought about adding `config-template-yarn.yml` or something so that a user could use

````{R}
spark_connect(master = "yarn-client", 
              config = 
                spark_config(
                  file = system.file(
                    file.path("conf", "config-template-yarn.yml"), 
                    package = "sparklyr"),
                  use_default = FALSE))
````

jjallaire:
For this use case it might make sense to have a `yarn` parameter on `spark_config` rather than requiring a reference to a file in the package.

MarcinKosinski:
So do you recommend to add `use_default_yarn = FALSE` parameter ? I could then provide a PR with `spark_config` extension`

jjallaire:
I'd recommend just `yarn = FALSE`. @javierluraschi What do you think?

MarcinKosinski:
semi PR is here https://github.com/rstudio/sparklyr/pull/142

MarcinKosinski:
moved to #142

PR #142, opened 2016-08-01
defaul config YAML for cluster deployment

MarcinKosinski:
proposition for #136 - default YARN.yyml configuration

### DIFF
--- a/R/config_spark.R
+++ b/R/config_spark.R
@@ -20,7 +20,8 @@ spark_config <- function(file = "config.yml", use_default = TRUE, yarn = FALSE) 
     localConfigFile <- system.file(file.path("conf", "config-template.yml"), package = "sparklyr")
     baseConfig <- config::get(file = localConfigFile)
   }
-
+  
+  yarnConfig <- list()
   if (yarn) {
     yarnConfigFile <- system.file(file.path("conf", "yarn-config-template.yml"), package = "sparklyr")
     yarnConfig <- config::get(file = yarnConfigFile)
--- a/R/config_spark.R
+++ b/R/config_spark.R
@@ -2,14 +2,18 @@
 #'
 #' @export
 #' @param file Name of the configuration file
-#' @param use_default TRUE to use the built-in detaults provided in this package
+#' @param use_default TRUE to use the built-in defaults provided in this package for local deplyomnet - \href{https://github.com/rstudio/sparklyr/blob/master/inst/conf/config-template.yml}{see}
+#' @param yarn TRUE to use the built-in defaults provided in this package for yarn cluster deplyomnet - \href{https://github.com/rstudio/sparklyr/blob/master/inst/conf/config-template.yml}{see}
 #'
 #' @details
 #'
 #' Read Spark configuration using the \pkg{\link[config]{config}} package.
 #'
 #' @return Named list with configuration data
-spark_config <- function(file = "config.yml", use_default = TRUE) {
+spark_config <- function(file = "config.yml", use_default = TRUE, yarn = FALSE) {
+  stopifnot(is.logical(yarn), length(yarn) == 1)
+  stopifnot(is.logical(use_default), length(use_default) == 1)
+
   baseConfig <- list()
 
   if (use_default) {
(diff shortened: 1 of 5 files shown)


R/config_spark.R
```diff
@@ -2,24 +2,33 @@
...
 
+  if (yarn) {
+    yarnConfigFile <- system.file(file.path("conf", "yarn-config-template.yml"), package = "sparklyr")
+    yarnConfig <- config::get(file = yarnConfigFile)
+  }
+
   userConfig <- tryCatch(config::get(file = file), error = function(e) NULL)
 
-  mergedConfig <- merge_lists(baseConfig, userConfig)
+  mergedConfig <- merge_lists(baseConfig, yarnConfig, userConfig)
```
jjallaire: Do we need `yarnConfig` to be set to `NULL` above for cases when `yarn = FALSE`?

R/config_spark.R
```diff
@@ -2,24 +2,33 @@
...
+#' @param use_default TRUE to use the built-in defaults provided in this package for local deplyomnet - \href{https://github.com/rstudio/sparklyr/blob/master/inst/conf/config-template.yml}{see}
+#' @param yarn TRUE to use the built-in defaults provided in this package for yarn cluster deplyomnet - \href{https://github.com/rstudio/sparklyr/blob/master/inst/conf/config-template.yml}{see}
 #'
 #' @details
 #'
 #' Read Spark configuration using the \pkg{\link[config]{config}} package.
 #'
 #' @return Named list with configuration data
-spark_config <- function(file = "config.yml", use_default = TRUE) {
+spark_config <- function(file = "config.yml", use_default = TRUE, yarn = FALSE) {
```
jjallaire: Are the yarn defaults so necessary/useful that we should find a way to pass `yarn = TRUE` automatically when the `master` starts with `yarn-`?

inst/conf/yarn-config-template.yml
```diff
@@ -0,0 +1,9 @@
+default:
```
jjallaire: If we are going to make these a default for `master = "yarn-*"` are these values truly universally desirable? (i.e. do they depend significantly on hardware or can they be taken for granted as a reasonable baseline for all clusters).

MarcinKosinski: @jjallaire I think if we'll go down with executor.instances to 5, then this is very universal setting and uses very few resources (if we would think about regular yarn cluster with +50 machines).

R/config_spark.R
MarcinKosinski: @jjallaire  Sure we need

R/config_spark.R
MarcinKosinski: As `spark_config` is used in `spark_connect` where there exists `master` parameter, maybe it could be

````{R}
spark_config <- function(file = "config.yml", use_default = TRUE, yarn = grepl("yarn", master))
````

for user interface simplification?

jjallaire:
Spent some time talking with the rest of the team about this and I think given the diversity of clusters and configurations users might encounter we are better off with enhancing the documentation around the use of config files (both generally and specifically with yarn). 

I think we'll want to emphasize here (http://spark.rstudio.com/deployment.html#cluster-deployment) that if you are connecting to a cluster you almost certainly will want to a config file (then provide an inline example). Further, we'll want a subsection on yarn clusters that provide your example config.

@nwstephens Is going to be looking at enhancing our deployment docs so will take the ball on making sure that it's extremely clear for someone connecting to yarn exactly how to do so.

MarcinKosinski:
Thanks for your effort guys! You are really doing staff that is necessary and will be used by many :) !

MarcinKosinski:
Last question if I may ask: how many people do you have in a team? This must be great to develop R packages in bigger group

jjallaire:
There are 3 of us working directly on the package and another 3 working on
examples, documentation, testing, etc.

MarcinKosinski:
Is this your regular amount of people during package development at RStudio or this project is more/less equipped with people :)?

jjallaire:
Depends on the package. This project is indeed staffed up quite a bit more
than others right now b/c there's so much work to do in order to get it
sufficiently robust and feature rich for production use (I think we're
getting close!).

MarcinKosinski:
I hope you'll succeed with this packages :) are your's TODOs perhaps listed on JIRA (or maybe on trello) or do you only use GitHub's issues?

jjallaire:
Just GitHub issues.

PR #146, opened 2016-08-03
merge sparkapi into sparklyr

kevinushey:
This PR merges the components of `sparkapi` back into `sparklyr`. The PR effectively:

1. Copies all of the `R` and `scala` files from `sparkapi` into `sparklyr`,
2. Renames all usages of `sparkapi` to `sparklyr`.

Do we want to rename any of the `sparkapi` functions here, e.g. should `invoke` become `spark_invoke`? We'll also need to update the website documentation to point back at `sparklyr` rather than `sparkapi`.


kevinushey:
I think this PR is good to go, but we should quickly chat about the line comments I've added before merge.

NAMESPACE
```diff
@@ -86,6 +105,12 @@ export(ft_quantile_discretizer)
 export(ft_sql_transformer)
 export(ft_string_indexer)
 export(ft_vector_assembler)
+export(hive_context)
```
kevinushey: I wonder if the methods here should become `spark_hive_context`, `spark_invoke`, and so on?

NAMESPACE
```diff
@@ -108,6 +133,9 @@ export(ml_random_forest)
 export(ml_save)
 export(ml_survival_regression)
 export(na.replace)
+export(print_jobj)
+export(register_extension)
```
kevinushey: Similarly, should this become `sparklyr_register_extension`?

It's a shame that there isn't a nice way to enforce certain functions must be called through `::` as `sparklyr::register_extension` is nice as an interface, but it's not necessarily a function we want to make available after attaching `sparklyr`.

jjallaire:
I don't think we should prefix all of these functions (e.g. "invoke", "register_extension", "hive_context", etc.). It's true that sparklyr does export a lot of functions, but they still seem very unlikely to overlap with other packages. This especially the case because the sparklyr workflow tends to be more self-contained (because other R packages aren't useful until you do a collect). I'll merge and then we can discuss later on today if we want to change those function names.

R/spark_shell.R
```diff
@@ -0,0 +1,387 @@
...
+                        config = list(),
+                        extensions = sparklyr::registered_extensions(),
+                        jars = NULL,
+                        packages = NULL,
+                        environment = NULL,
+                        shell_args = NULL) {
+  # read app jar through config, this allows "sparkr-shell" to test sparkr backend
+  app_jar <- spark_config_value(config, "sparklyr.app.jar", NULL)
+  if (is.null(app_jar)) {
+    app_jar <- shQuote(normalizePath(system.file(file.path("java", "sparklyr-1.6.1.jar"), package = "sparklyr"),
```
jjallaire: How do we manage to bind to sparklyr-2.0.0 when running under Spark 2.0?

PR #150, opened 2016-08-05
implement 'spark_compile_package_jars' for extension packages

kevinushey:
This PR encompasses some work that will make it easier for extension packages to compile + distribute jars powering Spark extensions.

The `sparklyr` package dogfoods itself here, calling `sparklyr::spark_compile_package_jars()` to build its own `sparklyr` jars.

# TODO

- [x] Use compiler version when building jar
- [x] Implement `spark_dependencies` as a mechanism for declaring what jars a package exposes


jjallaire:
I see that the scala compiler version is 2.11.8. Right now I'm passing 2.10 as the scala version to extensions b/c it's my understanding that the spark binary distributions are still built with scala 2.10

kevinushey:
I'll need to download Scala 2.10.6 so I can rebuid our jars with that.

kevinushey:
Also curious -- what's the main motivation behind using 'major.minor' over 'major.minor.patch' for the Spark version? Just wondering because when an extension is built, it's done with a specific version of Spark's jars on the classpath, so in theory it would be possible for a jar built against Spark 1.6.1 to be incompatible with Spark 1.6.2, although I imagine the Spark team would endeavor to avoid that.

R/sdf_sql.R
```diff
@@ -78,50 +78,6 @@ df_from_sql <- function(sc, sql) {
...
-        stringData,
-        fields,
-        rows)
-    })
-
-    df[[1]] <- NULL
-  }
-
-  df
+  sdf_collect(sdf)
```
javierluraschi: @kevinushey Could we do a quick check on performance here? `sdf_collect` seems to trigger one `collect()` per column, so for wide data frames, I would think this slows down performance. Is that the case? Or is the change in performance not even significant?

javierluraschi: I think a simple

```{r}
system.time({
   tbls("flights") %>% collect
})
```

with both versions would be sufficient here to give us an idea.

kevinushey: I should double check, but I believe the performance gains from reading columns as arrays of data offsets the cost over reading the entire data frame as an R list of lists (as this ends up requiring a lot more indirection in the deserialization process)

jjallaire:
My thinking is that we should only require a single JAR per major.minor
version (assuming there is in general an assurance of compatibility within
major.minor releases). Since we need to tell the extension which Spark
version to use and this version is used to construct the path then if we
passed "patch" versions would either have to:

a) Build a JAR for every major.minor.patch version to cover their bases; or

b) Do their own normalization of major.minor.patch to major.minor
internally.

By passing (and building) only against a major.minor version we bypass both
of these problems.

I could certainly be missing something though...

R/sdf_sql.R
javierluraschi: I believe even arrays need to be deserialized, currently, line by line: https://github.com/rstudio/sparklyr/blob/master/R/spark_deserialize.R#L113-L115. But yeah, a quick check to make sure this does not make things painfully slow, would be good.

jjallaire: I think the name of the function could just be "compile_package_jars"

kevinushey: @javierluraschi Ack, I'm glad you pointed this out -- it looks like my column reading routines were actually mis-handling NA values. (Although I think we can still read things a column at a time, I need to be more careful about handling NAs...)

I had assumed that, for an 'integer' column in a Spark data frame, all entries in that column would be encoded as integers. Unfortunately, this is not the case -- it looks like missing values are always encoded as logical values (?), so the read routines need to explicitly check that. Thanks!

kevinushey:
Bonus: this PR will come with much-improved `collect` performance!

Before:
```
> system.time(collected <- flights_tbl %>% collect())
   user  system elapsed 
129.575   0.999 135.212 
```

After:
```
> system.time(collected <- flights_tbl %>% collect())
   user  system elapsed 
  0.638   0.089   2.623 
```

A ~50x improvement doesn't seem half bad :)

R/spark_compile.R
```diff
@@ -126,7 +126,7 @@ spark_compile <- function(name,
 #'
 #' @keywords internal
 #' @export
-spark_compile_package_jars <- function(package = rprojroot::find_package_root_file(),
+compile_package_jars <- function(package = rprojroot::find_package_root_file(),
```
jjallaire: The devtools functions use pkg = "." which seems straightforward enough. Is the idea that people might be working in a package but not in the root directory?

R/spark_compile.R
```diff
@@ -126,7 +126,7 @@ spark_compile <- function(name,
...
                                        spark_versions = NULL,
```
jjallaire: I think we need to provide a default here which documents what we know is necessary (e.g. we know that moving from spark 1.6 to 2.0 likely requires a new JAR due to the change in type of DataFrame).

jjallaire: We might also want a scala_versions = c("2.10") here as well.

R/spark_compile.R
```diff
@@ -126,7 +126,7 @@ spark_compile <- function(name,
...
                                        scalac = NULL,
```
jjallaire: scalac = Sys.which("scalac")   ?

R/spark_compile.R
jjallaire: Since we target a single JAR with this function should we call it `compile_package_jar` ?

jjallaire:
I'm going to wait to merge this PR until Tuesday (significant enough change that some small problem might manifest and don't want any fire fighting to be necessary over the weekend).

R/spark_compile.R
kevinushey: We try to find the program using `path_program` when this is `NULL` below, although perhaps we should just make the default `path_program("scalac")` instead of `NULL`.

R/spark_compile.R
kevinushey: The tough part about scala_versions is that we don't really have a good way of indexing the `scala` compilers. Maybe we need an entry like:

```
scala_compilers = list(
    "2.10" = "path/to/scala-2.10",
    "2.11" = "path/to/scala-2.11"
)
```

or something to that effect?

R/spark_compile.R
jjallaire: But if path_program is one of our functions and is unexported (which it
seems like it should be) then I'd say it's better to keep it as NULL

R/spark_compile.R
kevinushey: `pkg` might be better as an argument name (for conformance with devtools), but I like using `rprojroot` since it makes it easier to call this function when you might have set the working directory to e.g. the subdirectory of a project.

E.g. if you needed to write a standalone R script that lived somewhere in the bowels of your package sources, this would still do the right thing without extra 'find my package root' work from the user.

R/spark_compile.R
jjallaire: Okay, let's leave the scala compiler alone and we'll just call the ones we
know need to be called (it can be a vectorized parameter so users can pass
multiple compilers).

R/spark_compile.R
jjallaire: Okay, let's leave this as-is then.

kevinushey: I could go either way on the function name as well (technically, we'll be generating one `jar` for each combination of Spark version * scala version provided)

jjallaire: Okay, let's leave it as jars then!

javierluraschi:
@kevinushey that's an awesome perfotmance boost! If you are interested, it would be good to compare agains sparkR collect. There are some examples on benchmarking sparkR in our docs.

jjallaire:
50x, awesome!!!!!  That is a game changer.

kevinushey:
Alright, I think this PR is in good shape now re: compilation of Spark extensions.

The default behavior of `compile_package_jars` now discovers:

- Spark 1.6.1 and Scala 2.10
- Spark 2.0.0 and Scala 2.11

and uses those during compilation. This implies most extension packages will only need:

```
sparklyr::compile_package_jars()
```

and things will just work. Users who need custom behavior can either do something like:

```
sparklyr::compile_package_jars(compilation_spec = list(
    spark_compilation_spec(
        spark_version = "1.6.1",
        scalac_path = "/path/to/scalac",
        jar_name = "package-1.6.jar"
))
```

Or, they can just manually call `spark_compile` by hand, passing the appropriate parameters.

@jjallaire let me know what you think / if we should tweak any names / interfaces here.

R/spark_compile.R
```diff
@@ -97,29 +81,174 @@ spark_compile <- function(name, spark_home) {
...
+#' @param jar The path to the \code{jar} Java archive tool to be used.
+#'   When unset, \code{jar} will be discovered on the PATH.
+#'
+#' @import rprojroot
+#' @import digest
+#'
+#' @keywords internal
+#' @export
+compile_package_jars <- function(package = rprojroot::find_package_root_file(),
+                                 compilation_spec = default_compilation_spec())
```
jjallaire: If we don't want to export `default_compilation_spec` we might just consider letting this be `"default"` and calling that function internally.

jjallaire: Should we consider automatically installing Spark 1.6.1 and/or Spark 2.0.0 if they aren't installed?

R/spark_compile.R
```diff
@@ -97,29 +81,174 @@ spark_compile <- function(name, spark_home) {
...
+
+    return(scalac_path)
+  }
+
+  stopf("failed to discover 'scalac %s' compiler", version)
+}
 
-  setwd(owd)
+get_scalac_version <- function(scalac = Sys.which("scalac")) {
+  # TODO: shell redirection won't work on Windows unless we go through shell
```
jjallaire: Let's add the shell call here.

R/spark_compile.R
jjallaire: Should we give an explicit error when we don't find versions of scalac required to meet the compilation spec?

kevinushey:
Okay, I think this is good to go.

@jjallaire, we do fail with an error if an appropriate Scala version is not discovered, e.g.

```
> library(sparklyr)
> sparklyr::compile_package_jars()
Error in find_scalac("2.11") : failed to discover 'scalac 2.11' compiler
Calls: <Anonymous> ... spark_default_compilation_spec -> spark_compilation_spec -> find_scalac -> stopf
```

Does that look okay?

jjallaire:
Yes, that looks great.

PR #164, opened 2016-08-10
implement 'collect' method for Dataset

kevinushey:
This PR implements a new Scala 'collect' routine that can be used to read an entire data set (ie, all of its columns) back into R in an expedited fashion. This should hopefully help mitigate issues related to slowness in the Spark scheduler.


jjallaire:
LGTM. @javierluraschi you can merge after review.

kevinushey:
Still have a couple more commits pending (want to clean up some of the older, no longer in-use APIs) and then I think we can merge.

kevinushey:
It's also nice that our work re: automagically generating different jars for the versions of Spark we support mean we can freely use the `DataFrame` class now, and don't necessarily need to create custom Scala code for different versions of Spark :)

kevinushey:
Okay, I think this is good to go now -- @javierluraschi can you review + merge if all looks well?

javierluraschi:
Yes, looks good!

PR #179, opened 2016-08-18
add `tol` parameter to kmenas to support tolerance of convergence

MarcinKosinski:
according to #178 

By the way
- fixed 1 typo in other function
- generated documentation for ml_evaluators

PR #180, opened 2016-08-18
concentration parameter for document and topics distributions in ml_lda

MarcinKosinski:
I am not quite sure what is the default value of `beta` and `alpha` as I can not find it here: https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.ml.clustering.LocalLDAModel


R/ml_lda.R
```diff
@@ -6,13 +6,21 @@
...
 #'
 #' @family Spark ML routines
 #'
+#' @note For terminology used in LDA model see \href{https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.ml.clustering.LDA}{Spark LDA documentation}.
+#'
 #' @export
 ml_lda <- function(x,
                    features = dplyr::tbl_vars(x),
                    k = length(features),
+                   alfa = 0.2,
```
kevinushey: Should this be `alpha`?

kevinushey:
One way to get Spark to spit out the default values for parameters (+ their documentation) is to call the `explainParams` method on the object; e.g.

```
lda <- invoke_new(sc, "org.apache.spark.ml.clustering.LDA")
params <- invoke(lda, "explainParams")
cat(params, sep = "\n")
```

For that, I see e.g.

```
> cat(params, sep = "\n")
checkpointInterval: set checkpoint interval (>= 1) or disable checkpoint (-1). E.g. 10 means that the cache will get checkpointed every 10 iterations (default: 10)
docConcentration: Concentration parameter (commonly named "alpha") for the prior placed on documents' distributions over topics ("theta"). (undefined)
featuresCol: features column name (default: features)
k: The number of topics (clusters) to infer. Must be > 1. (default: 10)
keepLastCheckpoint: (For EM optimizer) If using checkpointing, this indicates whether to keep the last checkpoint. If false, then the checkpoint will be deleted. Deleting the checkpoint can cause failures if a data partition is lost, so set this bit with care. (default: true)
learningDecay: (For online optimizer) Learning rate, set as an exponential decay rate. This should be between (0.5, 1.0] to guarantee asymptotic convergence. (default: 0.51)
learningOffset: (For online optimizer) A (positive) learning parameter that downweights early iterations. Larger values make early iterations count less. (default: 1024.0)
maxIter: maximum number of iterations (>= 0) (default: 20)
optimizeDocConcentration: (For online optimizer only, currently) Indicates whether the docConcentration (Dirichlet parameter for document-topic distribution) will be optimized during training. (default: true)
optimizer: Optimizer or inference algorithm used to estimate the LDA model. Supported: online, em (default: online)
seed: random seed (default: 1435876747)
subsamplingRate: (For online optimizer) Fraction of the corpus to be sampled and used in each iteration of mini-batch gradient descent, in range (0, 1]. (default: 0.05)
topicConcentration: Concentration parameter (commonly named "beta" or "eta") for the prior placed on topic' distributions over terms. (undefined)
topicDistributionCol: Output column with estimates of the topic mixture distribution for each document (often called "theta" in the literature).  Returns a vector of zeros for an empty document. (default: topicDistribution)
```

kevinushey:
...although apparently, in this case, the default values are undefined, so that doesn't really help! :)

MarcinKosinski:
Ok I've found the default valus for default EM optimization

>  topicConcentration: DoubleParam

Concentration parameter (commonly named "beta" or "eta") for the prior placed on topics' distributions over terms.

This is the parameter to a symmetric Dirichlet distribution.

Note: The topics' distributions over terms are called "beta" in the original LDA paper by Blei et al., but are called "phi" in many later papers such as Asuncion et al., 2009.

If not set by the user, then topicConcentration is set automatically. (default = automatic)

Optimizer-specific parameter settings:

    EM
        Value should be > 1.0
        default = 0.1 + 1, where 0.1 gives a small amount of smoothing and +1 follows Asuncion et al. (2009), who recommend a +1 adjustment for EM.
    Online
        Value should be >= 0
        default = (1.0 / k), following the implementation from https://github.com/Blei-Lab/onlineldavb.

Definition Classes
    LDAParams
Annotations
    @Since( "1.6.0" )

MarcinKosinski:
@kevinushey I've updated default values of beta and alpha to be the same as in Spark ML. 
I recommend to merge #179 before this PR can be merged

MarcinKosinski:
reopened again in https://github.com/rstudio/sparklyr/pull/186 due to too many conflicts

PR #190, opened 2016-08-25
Enables to get a SparkContext and to use a running R Backend

junyangq:
This PR tries to extend the flexibility of sparklyr in two aspects:
* Get rather than create a SparkContext if there is already one in the background
* Can connect to a running R backend if the user provides the port number

### DIFF
--- a/R/config_spark.R
+++ b/R/config_spark.R
@@ -3,6 +3,7 @@
 #' @export
 #' @param file Name of the configuration file
 #' @param use_default TRUE to use the built-in detaults provided in this package
+#' @param extraConfig List of configuration parameters provided by the user
 #'
 #' @details
 #'
--- a/R/config_spark.R
+++ b/R/config_spark.R
@@ -9,7 +9,7 @@
 #' Read Spark configuration using the \pkg{\link[config]{config}} package.
 #'
 #' @return Named list with configuration data
-spark_config <- function(file = "config.yml", use_default = TRUE) {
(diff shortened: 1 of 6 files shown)


javierluraschi:
@junyangq: We are currently in the process of submitting `sparklyr` to CRAN and therefore, we've closed api/config/documentation changes until the CRAN version is released a few weeks from today. We are planning on keeping this PR open and take this feedback after the CRAN release.

In the meantime, could you explain in which cases you are planning to manually specify the ports number? Ideally, users shouldn't need to deal with the ports number themselves; therefore, I would like to understand in which cases this is useful and explore if there are different ways to accomplish this.

junyangq:
@javierluraschi Thanks for your quick reply. It is common for businesses to have a running Spark cluster managed by some cluster manager. In many cases, the backend has already been started and multiple users may want to share the same SparkContext. So I think it would be great if sparklyr can support this mode. What do you think?

javierluraschi:
@junyangq appart from reusing the `SparkContext` we would need a way to allow multiple clients to connect to the sparklyr backend. We have a proposed change to support this multiple-client architecture under PR https://github.com/rstudio/sparklyr/pull/238. Therefore, planning to close this PR and address any feedback in https://github.com/rstudio/sparklyr/pull/238 instead, let us know your thoughts to iterate over them if needed.

kevinushey:
Thanks for the PR -- I think this would will effectively be subsumed by https://github.com/rstudio/sparklyr/pull/238, so I'm going to close this. Thanks!

PR #195, opened 2016-08-26
SQL Translation: modified `paste` to map to "CONCAT_WS" &  added `paste0` mapping to "CONCAT"

AlexIoannides:
Allows specification of a separator when contacting string variables, and brings behaviour more in-line with `paste`\`paste0` in base R.


javierluraschi:
We are about to checkpoint a CRAN release so we will take very few PR to make sure there is enough time to test the merges or be very confident they won't cause regressions. My initial thought is to merge this one before CRAN since: 1) it's safe, 2) provides functionality that would otherwise be hard to workaround and 3) changes the API which would be good to avoid after a CRAN release.

kevinushey:
LGTM as well, although there's unfortunately merge conflicts now.

AlexIoannides:
Do you need me to take a look at those?

kevinushey:
I've added this in by hand now -- thanks for the PR and sorry for not merging before conflicts arose!

Issue #211, opened 2016-09-11
pmin and pmax are unsupported


magarick:
Using pmin or pmax to get the elementwise minima or maxima of two columns does not work with a Spark backend. The following code works when using a plain R backend (and I believe most databases too).

`df_test <- data.frame(x=c(1, 2, 3), y=c(2, 1, 2))
df_test %>% mutate(pmin(x,y)) `

However, the following code with the same data in Spark does not.

`sc = spark_connect(master='local')
copy_to(sc, df_test)
sdf_load_table(sc, 'df_test') %>% mutate(pmin(x,y)) 
`

The error comes from Spark itself in Scala. The first line of the stack trace is
`Error: org.apache.spark.sql.AnalysisException: Invalid number of arguments for function min; line 2 pos 23`

With the help of some wiser Spark folks, I see that the min and max functions in Spark expect only one argument. Since the SQL generated by pmin and pmax uses MIN and MAX, I'm not sure where the fix should occur. Thanks for any help.

kevinushey:
We could probably add a function to support this. That said, 

```
df_tbl %>%
  mutate(z = ifelse(x < y, x, y))
```

should work okay now for the two variable case.

magarick:
Thanks! Solution works well.

PR #213, opened 2016-09-12
Implement pmin and pmax in dplyr interface

javierluraschi:
Suggestion from: https://github.com/rstudio/sparklyr/issues/211

### DIFF
--- a/R/dplyr_spark_connection.R
+++ b/R/dplyr_spark_connection.R
@@ -3,6 +3,43 @@ sql_escape_ident.spark_connection <- function(con, x) {
   sql_quote(x, '`')
 }
 
+build_sql_if_compare <- function(..., con, compare) {
+  args <- list(...)
+
+  build_sql_if_parts <- function(ifParts, ifValues) {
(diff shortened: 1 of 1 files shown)


R/dplyr_spark_connection.R
```diff
@@ -20,7 +57,9 @@ sql_translate_env.spark_connection <- function(con) {
       paste0 = function(...) build_sql("CONCAT", list(...)),
       xor = function(x, y) build_sql(x, " ^ ", y),
       or = function(x, y) build_sql(x, " or ", y),
-      and = function(x, y) build_sql(x, " and ", y)
+      and = function(x, y) build_sql(x, " and ", y),
+      pmin = function(...) build_sql_if_compare(..., con = con, compare = " <= "),
```
kevinushey: I think the whitespace around the operator here is unnecessary since you call `paste` above with ` and `

gbisschoff:
does pmin & pmax support na.rm? I get cannot resolve '(CAST(uzndxcbugq.`X` AS DOUBLE) <= true)

Issue #215, opened 2016-09-14
*_join(..., copy = T) doesn't work


kjedrzejewski:
I'm working on Spark 2.0.0 with local master. When i run following code:

``` R

n = 1000000
x = data.table(id = 1:n, val = rnorm(n), t = 1:10)
x_spark = sc %>% copy_to(x, overwrite = T)
ts = data.table(t = 1:10, name = letters[1:10])

x_spark %>%
  group_by(t) %>%
  summarize(m = mean(val)) %>%
  arrange(t) %>%
  inner_join(ts, by = c('t'), copy = T)

```

and i receive following error:

```
ERROR: org.apache.spark.sql.catalyst.parser.ParseException: 
CREATE TEMPORARY TABLE is not supported yet. Please use CREATE TEMPORARY VIEW as an alternative.(line 1, pos 0)

== SQL ==
CREATE TEMPORARY TABLE `rnprrlqxwk` (`t` INTEGER, `name` STRING)
^^^

    at org.apache.spark.sql.execution.SparkSqlAstBuilder$$anonfun$visitCreateTable$1.apply(SparkSqlParser.scala:905)
    at org.apache.spark.sql.execution.SparkSqlAstBuilder$$anonfun$visitCreateTable$1.apply(SparkSqlParser.scala:901)
    at org.apache.spark.sql.catalyst.parser.ParserUtils$.withOrigin(ParserUtils.scala:96)
    at org.apache.spark.sql.execution.SparkSqlAstBuilder.visitCreateTable(SparkSqlParser.scala:901)
    at org.apache.spark.sql.execution.SparkSqlAstBuilder.visitCreateTable(SparkSqlParser.scala:53)
    at org.apache.spark.sql.catalyst.parser.SqlBaseParser$CreateTableContext.accept(SqlBaseParser.java:474)
    at org.antlr.v4.runtime.tree.AbstractParseTreeVisitor.visit(AbstractParseTreeVisitor.java:42)
    at org.apache.spark.sql.catalyst.pa
```

which as I understand comes from `Spark` directly not `sparklyr`.

javierluraschi:
Thanks, sent a PR to implement joins against local tables, thanks for pointing this out: https://github.com/rstudio/sparklyr/pull/216

javierluraschi:
Forgot to close, we are good with this one:

```
> x_spark %>%
+   group_by(t) %>%
+   summarize(m = mean(val)) %>%
+   arrange(t) %>%
[...]

PR #216, opened 2016-09-15
add support for joins against local tables

javierluraschi:
This PR adds support for joining against local tables. The remaining work here was to allow dplyr to use `copy_to` to transfer the local table into Spark. See example from https://github.com/rstudio/sparklyr/issues/215:

```{r}
n = 1000000
x = data.table(id = 1:n, val = rnorm(n), t = 1:10)
x_spark = sc %>% copy_to(x, overwrite = T)
ts = data.table(t = 1:10, name = letters[1:10])

x_spark %>%
  group_by(t) %>%
  summarize(m = mean(val)) %>%
  arrange(t) %>%
  inner_join(ts, by = c('t'), copy = T)
```


javierluraschi:
We are holding off all PR, with the CRAN checkpoint to be performed (ETA) this Thursday. If someone needs this fix, one way yo get this is to use `devtools`:

```{r}
devtools::install_github(
   "https://github.com/rstudio/sparklyr",
   ref = "feature/join-local-tables")
```

Issue #114, opened 2016-07-17
spark_connect failed on OS X:Failed to launch Spark shell. Ports file does not exist.


feng5:
```
> library(sparklyr)
> library(dplyr)

Attaching package: ‘dplyr’

The following objects are masked from ‘package:stats’:

    filter, lag

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

> sc <- spark_connect(master = "local")
Error in sparkapi::start_shell(master = master, spark_home = spark_home,  : 
  Failed to launch Spark shell. Ports file does not exist.
    Path: /Users/13k/Library/Caches/spark/spark-2.0.0-preview-bin-hadoop2.6/bin/spark-submit
    Parameters: --jars, '/Users/13k/R/sparklyr/java/rspark_utils.jar', --packages, 'com.databricks:spark-csv_2.11:1.3.0','com.amazonaws:aws-java-sdk-pom:1.10.34', sparkr-shell, /var/folders/sd/2_0mf8p100v7dl3p_2cs5m_80000gn/T//RtmpXUplRl/file1323016779f8e.out
    
Ivy Default Cache set to: /Users/13k/.ivy2/cache
The jars for the packages stored in: /Users/13k/.ivy2/jars
:: loading settings :: url = jar:file:/Users/13k/Library/Caches/spark/spark-2.0.0-preview-bin-hadoop2.6/jars/ivy-2.4.0.jar!/org/apache/ivy/core/settings/ivysettings.xml
com.databricks#spark-csv_2.11 added as a dependency
com.amazonaws#aws-java-sdk-pom added as a dependency
:: resolving dependencies :: org.apache.spark#spark-submit-parent;1.0
	confs: [default]
Exception in thread "main" java.io.FileNotFoundException: /Users/13k/.ivy2/cache/resolved-org.apache.spar
```
sessioninfo:
```
> sessionInfo()
R version 3.3.1 (2016-06-21)
Platform: x86_64-apple-darwin13.4.0 (64-bit)
Running under: OS X 10.11.4 (El Capitan)

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] dplyr_0.5.0     sparklyr_0.2.24

loaded via a namespace (and not attached):
[...]

PR #222, opened 2016-09-20
revert use of readr to allow writeLines to reuse file connection while writ…

javierluraschi:
Revert https://github.com/rstudio/sparklyr/commit/ba8e2a to allow writeLines to reuse file connection while writing log settings.

Otherwise, users get the following warning during `spark_install`:

```
In value[[3L]](cond) : Failed to set logging settings
```

which was referenced in https://github.com/rstudio/sparklyr/issues/114

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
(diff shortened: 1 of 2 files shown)


javierluraschi:
I'll merge this one later today since it's a regression safe to apply and worth picking for the CRAN release.

PR #228, opened 2016-09-25
Add URL field to DESCRIPTION

gaborcsardi:


gaborcsardi:
Btw, it would be great if RStudio could generate a `DESCRIPTION` file that has `URL` and also `BugReports`. Thanks!

kevinushey:
You can get partway there by setting those in the `devtools.desc` option, e.g.

```
options(devtools.desc = list(
  URL = "path/to/repo",
  BugReports = "path/to/repo/issues"
))
```

PR #235, opened 2016-09-27
Add more Hive date support, espespecially a way to convert epoch time

ssoulier:


kevinushey:
Are these changes really necessary? One can already call these Hive functions when they are available. A silly example:

```
library(sparklyr)
library(dplyr)
sc <- spark_connect(master = "local", version = "2.0.0")

iris_tbl <- copy_to(sc, iris)
iris_tbl %>%
  mutate(date = current_date(),
         year = year(date),
         month = month(date),
         day = day(date))
```

The SQL transformations are really only necessary for R functions that one might normally call that need an explicit translation to an SQL equivalent.

ssoulier:
You are totally right.
My need was to convert epoch timestamp to date, for that I created `from_unixtime` and then I added the other date manipulation functions.
It was not clear for me that I can directly use HIVE function inside a mutate.
Could you show in the code where this magic is done ?

javierluraschi:
@ssoulier this magic is done somewhere in https://github.com/hadley/dplyr #tidyversemagic

PR #238, opened 2016-09-28
Replace ports file with gateway socket

javierluraschi:
This PR lays the ground work to provide more advanced connection models and removes the need to use an intermediate ports file to ease configuration across platforms.

# current connection architecture

sparklyr - backend improvements - crop

 - **spark-submit**: When executing spark_connect, sparklyr runs a shell command “spark-submit” which submits the `sparklyr` backend as a jar to Apache Spark..
 - **ports-file**: A binary file is written by the sparklyr backend to communicate back to sparklyr the sockets in the backend.
 - **socket**: Two sockets are provided, one is a monitor to destroy the backend once it closes, the other one requests java calls and serializes results back.

# improved connection architecture

This PR proposes using the monitor socket as a gateway socket that replaces the ports file and provides routing and reconnection to existing backends as follows:

sparklyr - new backend - crop

At this point, it is not intended to add or change functionality. Therefore, commands are expected to work without any additional change.

The only new configuration parameter worth mentioning is the default port the gateway service will attempt to use. In some very particular systems, if this port is already in use, please follow the following configuration pattern:

```{r}
config <- spark_config()

# default configuration settings for gateway socket
config[["sparklyr.gateway.port"]] <- "8880"
config[["sparklyr.gateway.address"]] <- "localhost"

sc <- spark_connect(master = "yarn-cluster", config = config)
```


jjallaire:
We might consider adding a `service` parameter directly to `spark_connect` in addition to being able to use the config file.

.classpath
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kevinushey: Do we want to commit this to the `sparklyr` repository?

.classpath
```diff
@@ -0,0 +1,10 @@
...
+<classpath>
+	<classpathentry kind="src" path="java"/>
+	<classpathentry kind="lib" path="inst/java/sparklyr-1.6-2.10.jar"/>
+	<classpathentry kind="lib" path="inst/java/sparklyr-2.0-2.11.jar"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
+	<classpathentry kind="con" path="org.scala-ide.sdt.launching.SCALA_CONTAINER"/>
+	<classpathentry kind="lib" path="/Users/javierluraschi/Library/Caches/spark/spark-1.6.2-bin-hadoop2.6/lib/spark-assembly-1.6.2-hadoop2.6.0.jar"/>
```
kevinushey: If we want this in the repo, is it possible to declare this as a relative path?

.project
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kevinushey: Same here -- should we have this in the repo / add to `.Rbuildignore`?

R/spark_shell.R
```diff
@@ -120,79 +134,110 @@ start_shell <- function(master,
...
-    else {
-      system2(spark_submit_path,
-              args = shell_args,
-              stdout = output_file,
-              stderr = output_file,
-              wait = FALSE)
+  }
+
+  tryCatch({
+    readInt <- function(con, n = 1) {
```
kevinushey: We already have `readInt` defined in `sparklyr`; do we need to redefine it here?

R/spark_shell.R
```diff
@@ -120,79 +134,110 @@ start_shell <- function(master,
...
-    ), {
-      stop(paste(
-        "Failed to launch Spark shell. Ports file does not exist.\n",
-        "    Path: ", spark_submit_path, "\n",
-        "    Parameters: ", paste(shell_args, collapse = ", "), "\n",
-        "    \n",
-        paste(readLines(output_file), collapse = "\n"),
-        if (file.exists(error_file)) paste(readLines(error_file), collapse = "\n") else "",
-        sep = ""))
+    writeInt <- function(con, value) {
```
kevinushey: This should already be defined as well.

java/backend.scala
```diff
@@ -70,56 +75,214 @@ class Backend {
...
+    }
+    
+    hc
+  }
+  
+  def portIsAvailable(port: Int) = {
+    var ss: ServerSocket = null
+    var available = false
+    
+    Try {
```
kevinushey: Does `Try` just swallow exceptions? Should we be catching / logging on failure?

java/backend.scala
```diff
@@ -70,56 +75,214 @@ class Backend {
...
+        ss = new ServerSocket(port, 1, InetAddress.getByName("localhost"))
+        available = true
+    }
+    
+    if (ss != null) {
+        Try {
```
kevinushey: Similar logging here?

java/backend.scala
```diff
@@ -70,56 +75,214 @@ class Backend {
...
         }
-      }.start()
-
-      backend.run()
+      }
+      
+      gatewayServerSocket.setSoTimeout(0)
+    
+      while(true) {
+        bind()
```
kevinushey: I'm probably not understanding exactly what's going on here, but it looks like we're repeatedly calling `bind()` within an infinite loop, and that call to bind unconditionally launches a number of threads. Can you elaborate a bit on what's going on here?

.classpath
javierluraschi: @kevinushey what's your take on this? I definitely had to use Eclipse to develop this set of changes so I'm planning to keep the Eclipse project in my working directory; however, in the short term, you are probably the only other person making changes to the Scala code. So, have you been using Eclipse already for your Scala changes? If not, would having an Eclipse project be helpful to you? If the answer is no, then I don't think it's worth checking this in, otherwise, I'll parameterize the paths to make you can use this as well.

javierluraschi: Spent some time here, unfortunately, eclipse is not great at parameterizing `.classpaths`. They recommend maven building the `.classpath` file, etc. Removing for now.

.classpath
javierluraschi: Removed.

.project
javierluraschi: I still think is helpful, we would still have to add classpaths manually to the project though.

R/spark_shell.R
javierluraschi: Thanks. Removed.

R/spark_shell.R
javierluraschi: Removed.

java/backend.scala
javierluraschi: Yes, Try swallow exceptions, that was my intention since we are trying to discover is a port is open.

java/backend.scala
javierluraschi: Same, by design swallowing exceptions.

java/backend.scala
javierluraschi: @kevinushey Good point! Wrote this instead to describe in details what is going on: https://github.com/rstudio/sparklyr/blob/feature/service-socket/java/backend.scala#L24-L68

kevinushey: Awesome, thanks! Much clearer now :)

PR #240, opened 2016-09-28
use ASCII unit separator for data copy

kevinushey:
This PR switches from using `|` as a delimiter to the (non-printable) ASCII unit separator. This should help resolve issues where attempts to copy data containing `|` as fields could fail.


javierluraschi:
LGTM!

PR #246, opened 2016-10-04
Correct definition of default "quote" argument

slygent:
Just a silly typo in the documentation.


kevinushey:
Whoops, thank you very much!

PR #257, opened 2016-10-06
implement handling for 'na.action'; default to 'na.omit'

kevinushey:
This PR implements handling for a new `ml.option` parameter, `na.action`. This parameter is either an R function, or the name of an R function, and its job is to take a Spark DataFrame, and return a transformed Spark DataFrame (e.g. by dropping missing values, or replacing missing values with 0, and so on.)

This PR also makes the default behavior of `na.action = na.omit`, so that by default missing values are dropped. This change might be contentious, but it should be a good change overall as most Spark.ML methods don't function well with missing values.

We might want to report something like how many values were dropped in such a case, however.


javierluraschi:
looks good!

PR #258, opened 2016-10-06
implement livy connections

javierluraschi:
This PR implements `livy` connections in `sparklyr` by starting and making use of an interactive Scala Livy session. The basic use of `livy` in `sparklyr` is as follows:

```{r}
library(sparklyr)
livy_install()
livy_service_start()

sc <- spark_connect(master = "local", method = "livy")
copy_to(sc, iris)
src_tbls(sc)

spark_disconnect(sc)
livy_service_stop()
```

**Background**: Livy supports Scala, Python and R sessions; with the R sessions being served by an `sparkR` client (https://github.com/cloudera/livy#session-kind). We could, potentially, send a PR to use `sparklyr` as a type of session to Livy; however, we would still have to write an `sparklyr` client to serialize into the remote Livy `sparklyr` session running in Livy which adds additional complexity compared to using `sparklyr` as a client to an Scala interactive session and therefore, this approach was taken.

**Design**: This implementation reuses most of the `sparklyr` backend in Livy. The motivation behind reusing most of the backend code is that, otherwise:
 1. Would require to re-implement method signature-matching that the `sparklyr` backend supports. For instance, `invoke(jobj, list("a"))` or `invoke(jobj, c("a"))` is mapped to an scala type with the correct conversion; there are other more complicated mappings involving maps and other structures.
 2. Would require to re-implement serialization to match the `sparklyr` backend. For instance, fetching an Scala nested array `Array[Array[], Array[]]` correctly into a `list(list(), list())` requires custom serialization.

The `sparklyr` backend is reused by pre-processing the `scala` sources into livy-compatible sources that are compiled using `livy_sources_refresh()` and placed for distribution under `inst/livy/*.scala`. When a livy session starts, those files are sourced into the livy session. Once the `sparklyr` backend is loaded, instead of using a socket to serialize binary data to the backend, serialization is performed in`sparklyr` by executing `livy` commands in base64, for instance, `sparklyr` would send the following `livy` statement to initialize the connections:

```{scala}
var sparklyrRetVar_0 = LivyUtils.invokeFromBase64("AAAAHm9yZy5hcGFjaGUuc3BhcmsuU3BhcmtDb250ZXh0AAAAAAEAAAAMZ2V0T3JDcmVhdGUAAAAAAA=="
```

In the remote Scala session through `livy`, `LivyUtils.invokeFromBase64` would deserialize this call and map it to the `sparklyr` backend, get the binary response and encode it back as base64 to which livy returns a `text/plain` response of:

```
sparklyrRetVar_0: String = AAAAAGoAAAABMA==
```

This response is now parsed in the `sparklyr` client and the base64 decoded back to a binary response. In cases where the response is too big and gets truncated, a livy magic is used to return this as json, for instance, by executing the following `livy` statement:

```
%json sparklyrRetVar_0
```

Once the serialization is in place, the rest of `sparklyr` works as expected and no fixes had to be performed in higher layers of `sparklyr`, say: `dplyr`, `ml` nor `extensions`.

**Known Issues**:
  1. Serialization of large datasets through `copy_to` is currently failing in Livy, investigation to mitigate this underway.


kevinushey:
Do we plan on adding something like `livy_install`, analogous to `spark_install`, for easy installation + use of `livy`?

javierluraschi:
Adding `livy_install` seems reasonable. The install entails downloading this: http://archive.cloudera.com/beta/livy/livy-server-0.2.0.zip, running `SPARK_HOME=<spark_home> ./livy-server` and to kill the process to stop. We would need also `livy_start` and `livy_stop`. For windows is a bit more work since they currently only provide a launcher for `bash`: https://github.com/cloudera/livy/blob/master/bin/livy-server but it wouldn't be a stretch to provide something similar for windows. I'll work on this after finishing the core integration to make sure we don't run out of time.

kevinushey:
I can take a shot at `install_livy`.

R/data_copy.R
```diff
@@ -65,7 +65,7 @@ spark_serialize_csv_string <- function(sc, df, columns, repartition) {
   }), optional = TRUE)
 
   tempFile <- tempfile(fileext = ".csv")
-  write.table(df, tempFile, sep = "\31", col.names = FALSE, row.names = FALSE, quote = FALSE)
```
javierluraschi: `\31` is not a valid json character that we use in livy, need to take a look a this a bit more and scope to this fix to livy only or find a valid separator for both, sockets and json.

R/precondition.R
```diff
@@ -132,16 +132,17 @@ require_directory_exists <- function(path, fmt = NULL) {
   path
 }
 
-ensure_directory <- function(path, ...) {
+ensure_directory <- function(path) {
```
javierluraschi: ensure_directory was only used in `spark_compile` now in `livy_install` as well.

R/spark_connection.R
```diff
@@ -213,51 +208,10 @@ print.spark_web_url <- function(x, ...) {
 }
 
 initialize_connection <- function(sc) {
-  sc$spark_context <- invoke_static(sc, "sparklyr.Backend", "getSparkContext")
```
javierluraschi: This code is now subclassed in `spark_shell_connection.initialize_connection`

java/utils.scala
```diff
@@ -213,7 +213,7 @@ object Utils {
     partitions: Int): RDD[Row] = {
 
     var data = rows.map(o => {
-      val r = o.split("\31", -1)
+      val r = o.split("\\|~\\|")
```
kevinushey: The `-1` argument is necessary here as otherwise the final record on a line could be dropped, e.g. as in `foo,bar,baz,`.

R/data_copy.R
javierluraschi: scoped to livy only

java/utils.scala
javierluraschi: Also, kept the `\31` character and scoped the fix to `livy` only.

PR #259, opened 2016-10-07
implement 'livy_install'

kevinushey:
This PR implements `livy_install()`. We download a version of livy (currently, only 0.2.0 is available), and then install it into the cache directory.

Each version of livy installed will be associated with a unique version of Spark, so a different livy installation is needed for each version of Spark used -- I'm not yet sure if we can relax that requirement.


R/install_livy.R
```diff
@@ -0,0 +1,70 @@
...
+#' \samp{livy} provides a REST API to Spark.
+#'
+#' @param spark_home The path to a Spark installation. The downloaded and
+#'   installed version of \samp{livy} will then be associated with this Spark
+#'   installation.
+#' @param livy_version The version of \samp{livy} to be installed. Currently,
+#'   only the beta version, 0.2.0, is available.
+#' @param hadoop_conf_dir Optional; the path to a Hadoop configuration directory.
+#' @export
+livy_install <- function(spark_home = Sys.getenv("SPARK_HOME"),
```
javierluraschi: I think we want to replace `spark_home` for a `version` parameter to specify the spark version. When null, we can get the `spark_home` using `spark_home_dir()` instead and also pass the version from the parameter down as `spark_home_dir(version = version)`. That way, users don't need to specify a `SPARK_HOME` explicitly. I think a `version` parameter is better since we would also want to check against supported Spark versions, see comment bellow. Other than that, this looks great!

R/install_livy.R
```diff
@@ -0,0 +1,70 @@
...
+                         livy_version = "0.2.0",
+                         hadoop_conf_dir = NULL)
+{
+  hadoop_conf_dir <- ensure_scalar_character(hadoop_conf_dir, default = "")
+
+  # ensure that spark home exists
+  if (!file.exists(spark_home))
+    stopf("No Spark installation found at '%s'", spark_home)
+
+  # use the directory path as the version
```
javierluraschi: We should add a check, since I just verified that Livy `0.2.0` does not support Spark `2.0.0`. See this mail thread: https://groups.google.com/a/cloudera.org/forum/#!topic/livy-dev/50qlsi-g23I.

R/install_livy.R
kevinushey: Good idea, thanks!

R/install_livy.R
kevinushey: Yikes, okay. I wonder if we should try to auto-select a Spark version based on the livy version requested?

javierluraschi: Some good news... just noticed they just bumped the version to `0.3` and took this PR to support `2.0.0`: https://github.com/cloudera/livy/pull/186 - Therefore, they are probably not far away from pushing `0.3`. They do claim it is backwards compatible with `1.6` but I think it's still great that this will let users choose their `livy` version in case they need to revert.

R/install_livy.R
```diff
@@ -0,0 +1,93 @@
...
+#'   installed version of \samp{livy} will then be associated with this Spark
+#'   installation. When unset (\samp{NULL}), the value is inferred based on
+#'   the value of \samp{spark_version} supplied.
+#' @param spark_version The version of Spark to use. When unset (\samp{NULL}),
+#'   the value is inferred based on the value of \samp{livy_version} supplied.
+#'   A version of Spark known to be compatible with the requested version of
+#'   \samp{livy} is chosen when possible.
+#' @param livy_version The version of \samp{livy} to be installed.
+#' @export
+livy_install <- function(spark_home    = NULL,
```
javierluraschi: Users usually set `SPARK_HOME` using `Sys.setenv(SPARK_HOME = "")` or directly in their environment. Therefore, I think we can remove this parameter and use `Sys.getenv("SPARK_HOME")` bellow. For `spark_version`, JJs preference suggestion has been to not name `spark_version` but rather only `version` and we've assumed this through the code so I would also rename this one to `version`.

kevinushey: But `version` as a placeholder for the Spark version makes sense in the context of Spark functions -- here, I think users would infer that `version` is associated with the `livy` version to be installed.

For an installation function, IMHO it makes more sense to request the user either explicitly provide a Spark version (in which case we'll construct a Spark home path for them), or a whole Spark home (in which case we assume the user knows what they're doing)

javierluraschi:
@kevinushey gave this a shot and merged this to the feature/livy branch. Found out that while using `livy_install` this file is missing exec permissions:

```
Javiers-iMac:bin javierluraschi$ ls -l
total 16
-rw-r--r--  1 javierluraschi  staff  4298 Oct 18 10:02 livy-server
```

Which does not happen while manually downloading the livy server file. I don't remember having this issue with `spark_install` so I wonder if this rings any bells.

BTW. No need to rush this since we are currently planning to slowdown the development of livy for `sparklyr` since we've found a couple issues that need to be fixed in `livy` first (performance, support for 2.0.0, authentication documentation, etc) so this work is going to stay in a branch for a while.

javierluraschi:
I merged your work into the `feature/livy` branch, closing this one.

PR #283, opened 2016-10-25
Commentd out shell_args <- NULL so that the shell args will be passed to the command line

kagren:
shell_args was always set to NULL. This had the undesired effect that the "sparklyr.shell.xxxxx" config settings were always ignored.


javierluraschi:
@kagren can you remove this rather than comment?

kagren:
@javierluraschi done

Issue #261, opened 2016-10-10
Columns parameter not available in spark_read_csv


rvm-xx:
Call from spark_read_csv to spark_csv_read (line 74 from data_interface.R) does not allow providing a schema for interprating columns.
Would adding an extra "columns" parameter to spark_csv_read with NULL as default and passing it to spark_read_csv allow sending a schema for reading the file?

javierluraschi:
@rvm-xx implemented.

PR #289, opened 2016-10-27
add support for infer_schema and columns options to spark_read_csv

kykuo:
Expose `infer_schema` and `columns` options to `spark_read_csv`. The former is defaulted to `FALSE` even if no schema is provided because the inference takes a long time for large datasets.

Related to https://github.com/rstudio/sparklyr/issues/261

### DIFF
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -34,7 +34,7 @@ spark_csv_options <- function(header,
 #'   Defaults to \code{TRUE}.
 #' @param columns A named vector specifying column types.
 #' @param infer_schema Boolean; should column types be automatically inferred?
-#'   Requires one extra pass over the data. Defaults to \code{FALSE}.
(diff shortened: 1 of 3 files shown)


R/data_interface.R
```diff
@@ -58,6 +63,8 @@ spark_read_csv <- function(sc,
                            name,
                            path,
                            header = TRUE,
+                           columns = NULL,
+                           infer_schema = FALSE,
```
javierluraschi: `infer_schema` needs to be set to `TRUE` to maintain compatibility with the existing API.

R/data_interface.R
```diff
@@ -1,4 +1,5 @@
 spark_csv_options <- function(header,
```
javierluraschi: There are two other calls to `spark_csv_options` that require to pass `TRUE` as the second parameter.

PR #301, opened 2016-10-31
initial support for (serial) 'do'

kevinushey:
This PR implements support for `dplyr::do`, albeit currently only in the 'serial' way.

The main caveat with this implementation is that it always returns an R dataset (e.g. a `tbl_df`) rather than a `tbl_spark`.

Some more work is needed to ensure that we properly handle the use of `.` arguments as `dplyr::do_` does.

I'm hoping that in the future the 'parallel' implementation of this function could act as a drop-in replacement for what we have here, and users could select whether they desire a serial vs. parallel implementation.

TODO before merge:

- [x] Unit tests
- [ ] Handle missing `.` in `do` expression


ElianoMarques:
@kevinushey pretty cool!

Issue #171, opened 2016-08-15
Would you please publish the JAR that comes with this R package as a Spark Package


falaki:
It will make installation of the R package easier in other Spark environments especially with regards to class path management. 

Spark Packages website: https://spark-packages.org/
Instructions on how to register a package: https://spark-packages.org/register/explain

jjallaire:
We'll consider doing this when we publish a stable version of the package
to CRAN (should be in the next 2 or 3 weeks)

falaki:
Is there still interest in this?

jjallaire:
Yes, getting the deployment topology right in all environments is our top
focus right now. Interestingly, we've heard from other
customers/environments that any use of Spark packages is problematic, see:
https://github.com/rstudio/sparklyr/pull/314

I think we'll need to do both and allow users to configure the behavior via
a global option.

@javierluraschi
<https://github.com/rstudio/rstudio/commits/master?author=javierluraschi>
@kevinushey

javierluraschi:
Just submitted `sparklyr` for verification to https://spark-packages.org/ (which seems to be the first step), I'll keep this thread posted. The plan would be to submit the CRAN version `0.4` and update them both (CRAN and https://spark-packages.org/) as new releases of `sparklyr` become available. @jjallaire

PR #314, opened 2016-11-09
Embed csv parser and dependencies directly into sparklyr

javierluraschi:
Several users have experienced problems installing the databricks' csv package from https://spark-packages.org/ due to local connectivity issues or security restrictions in closed-cluster environments.

This change embeds the csv package into `sparklyr` and it's cone dependency.


jjallaire:
Interesting juxtaposition between this PR and this issue: https://github.com/rstudio/sparklyr/pull/314

One saying "please make everything a Spark package for ease of deployment" and the other saying "please make nothing a Spark package for ease of deployment".

I wonder if there is a way for us to support both approaches via a global option.

R/spark_shell.R
```diff
@@ -1,4 +1,11 @@
+shell_connection_validate_config <- function(config) {
+  if ("spark.jars.default" %in% config) {
```
hadley: Shouldn't that be `names(config)`?

javierluraschi: Yes!

javierluraschi:
@jjallaire currently, if someone doesn't want the csv package, they could remove this form the `config.yml` file, I wonder if that's enough as a global option.

Is https://github.com/rstudio/sparklyr/issues/315 the issue you are referring to? One interesting thing to take a look at, is as to why the R session crashes when the java process crashes. Haven't taken a look at this yet but maybe this already sounds familiar? We start the java process through a bash script running through `system2`, I would expect that if this crashes, it has no effect on the R session, but perhaps this is caused by the socket connection terminating unexpectedly or something else.

javierluraschi:
Ah, this issue https://github.com/rstudio/sparklyr/issues/171! Yes, that's interesting since the argument there would be to pull everything out of `sparklyr`, as you mention, we probably need both paths.

falaki:
Note that CSV data source has been inlined in Spark 2.x. Including csv Jar file with spark 2.x will cause class loader problems.

inst/conf/config-template.yml
```diff
@@ -4,9 +4,13 @@ default:
   spark.sql.shuffle.partitions.local: !expr parallel::detectCores()
   spark.env.SPARK_LOCAL_IP.local: 127.0.0.1
 
+  # default embedded packages to load
+  sparklyr.jars.default:
+    - !expr system.file(file.path("java", "spark-csv_2.11-1.3.0.jar"), package = "sparklyr")
+    - !expr system.file(file.path("java", "commons-csv-1.1.jar"), package = "sparklyr")
```
falaki: You may also want to include uniVocity-parsers jar file. spark-csv supports an option to chose parser library. Some features only work with univocity. Also there is slight difference in parsing behavior.

javierluraschi: Good point, added.

javierluraschi:
Thanks @falaki, I was not aware that the `spark-csv` package is now included in Spark 2.0. `sparklyr` will default to use the inline version from Spark 2.0 onwards, otherwise, the embedded version will be deployed by default.

javierluraschi:
Planing to merge this soon.

PR #324, opened 2016-11-14
set sparklyr.do.implicit.dot to FALSE

javierluraschi:
@kevinushey:

I think setting this to `TRUE` makes this feature a bit harder to use, for instance, I was helping someone troubleshoot:

```
iris_tbl %>% group_by(Species) %>% do(other = "x")
```

The other user we know of was following this guide: https://cran.r-project.org/web/packages/dplyr/dplyr.pdf which also uses explicit `.` so I can't find a good example at the moment where people would expect to pass `.` implicitly, thoughts?

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -8,5 +8,5 @@ set_option_default <- function(...) {
 
 .onLoad <- function(...) {
   set_option_default(sparklyr.na.action.verbose = TRUE,
-                     sparklyr.do.implicit.dot   = TRUE)
+                     sparklyr.do.implicit.dot   = FALSE)
 }


kevinushey:
The intention of that option was to support syntax of the form:

```
  S <- diamonds_tbl %>%
    group_by(color, clarity) %>%
    do(model = ml_linear_regression(price ~ x + y + z))
```

But I'm thinking this was probably misguided so maybe we should just remove that option altogether. We should likely amend the `ml_` interfaces so that the following works:

```
  S <- diamonds_tbl %>%
    group_by(color, clarity) %>%
    do(model = ml_linear_regression(price ~ x + y + z, data = .))
```

(or something similar?)

kevinushey:
I'm just going to remove the implicit do stuff on mater -- we can close this PR.

Issue #244, opened 2016-10-02
Spark DF column limit? 


cam333:
Hi all, 
I've been playing around with sparklyr and seem to be running into a column limit of some sort?  I have a data set that has around 4500 columns and couldn't get it into a spark DF, but subsets of the df work.  
Here is an example:

Works:
``` 
library(sparklyr)
library(dplyr)
sc <- spark_connect(master =  "local")
df <- matrix(0, ncol = 200, nrow = 2) %>% as_data_frame()
sdf <- copy_to(sc, df, "df", overwrite = TRUE)

```

When I up the columns to 5000 I get an error:

``` 
library(sparklyr)
library(dplyr)
sc <- spark_connect(master =  "local")
df <- matrix(0, ncol = 5000, nrow = 2) %>% as_data_frame()
sdf <- copy_to(sc, df, "df", overwrite = TRUE)

```

The following error is thrown:

```
Error in if (ncol(df) > 0) df else invisible(df) : 
  argument is of length zero
In addition: Warning message:
In value[[3L]](cond) : java.lang.StackOverflowError
	at scala.collection.IndexedSeqOptimized$class.foreach(IndexedSeqOptimized.scala:32)
	at scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:34)
	at scala.collection.generic.Growable$class.$plus$plus$eq(Growable.scala:48)
	at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:176)
	at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:45)
	at scala.collection.generic.GenericCompanion.apply(GenericCompanion.scala:48)
	at org.apache.spark.sql.catalyst.expressions.BinaryExpression.children(Expression.scala:358)
	at org.apache.spark.sql.catalyst.trees.TreeNode.containsChild$lzycompute(TreeNode.scala:86)
	at org.apache.spark.sql.catalyst.trees.TreeNode.containsChild(TreeNode.scala:86)
	at org.apache.spark.sql.catalyst.trees.TreeNode$$anonfun$4.apply(TreeNode.scala:280)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:328)
	at scala.collection.Iterator$class.foreach(Iterator.scala:727)
	at scala. [... truncated]
```

Any idea what's going on? 

Thanks

kevinushey:
[...]

PR #327, opened 2016-11-16
use friendly names when using copy_to

javierluraschi:
@kevinushey While fixing #244, with https://github.com/rstudio/sparklyr/commit/b23b97d90a43514d245e90ba06abe00c7a411420 we avoided an Spark bug in `< 2.0.0` by not using the `CACHE TABLE` statement from SQL. This had the unintended consequence of displaying table names as:

screen shot 2016-11-15 at 4 12 09 pm

While we don't have a better workaround for this Spark issue, this change scopes the workaround to only tables with 1000+ columns (verified that 1000 columns does not trigger the stackoverflow in spark) to display tables as:

screen shot 2016-11-15 at 4 12 15 pm

### DIFF
--- a/R/dplyr_spark.R
+++ b/R/dplyr_spark.R
@@ -113,6 +113,22 @@ copy_to.src_spark <- function(dest, df, name, ...) {
   copy_to(spark_connection(dest), df, name, ...)
 }
 
+tbl_cache_sdf <- function(sc, name, force) {
+  tbl <- tbl(sc, name)
+  sdf <- spark_dataframe(tbl)
+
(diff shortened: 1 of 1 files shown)


kevinushey:
Would it be possible to assign the table some temporary, randomly-generated name before calling cache instead?

javierluraschi:
@kevinushey I don't think so, this UI shows the RDD name, not the Hive table name which is set correctly without this fix. There might be some other way, so I'll dig a bit further; but if not, would you have concerns merging this one? I do think this is ugly enough to try to fix it for `0.5`

PR #334, opened 2016-11-23
Refactor copy_to into sdf_import

javierluraschi:
This is a merge from the functionality in `copy_to` (the copying based on serialization methods to support cluster copies) and the functionality in `sdf_import` (the hashing and conversion to data frame if needed).

Two reasons fopr this change:
1) Is confusing to users that only `copy_to` is redocumented in the `dplyr` documentation. Instead, we want `sdf_copy_to` to be the default entry point for documentation since there are users that use `sparklyr` but not `dplyr`.
2) `sdf_import` does not support copying into a cluster where the HDFS file that is persisted is not available to all nodes, therefore, we use the `string_serializer` as needed in `sdf_import` as well.

I'm also planning to remove `copy_to` from the documentation and rename the "dplyr Interface" section into "dplyr Extensions" since we want to not re-document each dplyr function.

CC: @kevinushey

### DIFF
--- a/R/data_copy.R
+++ b/R/data_copy.R
@@ -9,8 +9,20 @@ spark_data_build_types <- function(sc, columns) {
 }
 
 spark_serialize_csv_file <- function(sc, df, columns, repartition) {
-  tempfile <- tempfile(fileext = ".csv")
-  write.csv(df, tempfile, row.names = FALSE, na = "")
+
+  # generate a CSV file from the associated data frame
+  # note that these files need to live for the R session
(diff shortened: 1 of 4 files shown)


kevinushey:
Is that hashing + writing to disk stuff even needed now? (It looks like the serializer chosen won't ever use that information).

javierluraschi:
@kevinushey Thanks for pointing this out! We can still use the hash, but only on the CSV file serializer, move this there.

PR #335, opened 2016-11-23
Changes to tbl_cache and sdf_read/write

javierluraschi:
1. Refactored `tbl_cache` and `tbl_uncache` to not use `dplyr`. This is to allow the non-dplyr features in `sparklyr`  to not require `dplyr`.
 2. Updated documentation to simplify reference guide. For instance, removed `na.replace` and would argue that `dim`, `ncol`, etc. should not be documented since, in the same way that `dplyr` and `DBI` interfaces are not explicitly documented, we shouldn't duplicate documentation. Thoughts welcomed.
 3. Deprecated `sdf_read` and `sdf_write` and provided instead `spark_read` and `spark_write` equivalents to maintain consistency.

CC: @kevinushey


R/sdf_saveload.R
```diff
@@ -16,6 +16,8 @@ NULL
 #' @rdname sdf-saveload
 #' @export
 sdf_save_table <- function(x, name, overwrite = FALSE, append = FALSE) {
+  warning("sdf_save_table is deprecated, use spark_write_table instead.")
```
kevinushey: You can use `.Deprecated("spark_write_table")` instead here.

R/sdf_saveload.R
```diff
@@ -39,6 +41,8 @@ sdf_save_table <- function(x, name, overwrite = FALSE, append = FALSE) {
 #' @rdname sdf-saveload
 #' @export
 sdf_load_table <- function(sc, name) {
+  warning("sdf_load_table is deprecated, use spark_read_table instead.")
```
kevinushey: `.Deprecated("spark_read_table")`?

R/sdf_saveload.R
```diff
@@ -60,6 +64,8 @@ sdf_load_table <- function(sc, name) {
 #' @rdname sdf-saveload
 #' @export
 sdf_save_parquet <- function(x, path, overwrite = FALSE, append = FALSE) {
+  warning("sdf_save_parquet is deprecated, use spark_write_parquet instead.")
```
kevinushey: `.Deprecated("spark_write_parquet")`?

R/sdf_saveload.R
```diff
@@ -73,6 +79,8 @@ sdf_save_parquet <- function(x, path, overwrite = FALSE, append = FALSE) {
 #' @rdname sdf-saveload
 #' @export
 sdf_load_parquet <- function(sc, path) {
+  warning("sdf_load_parquet is deprecated, use spark_read_parquet instead.")
```
kevinushey: `.Deprecated("spark_read_parquet")`

R/tables_spark.R
```diff
@@ -0,0 +1,73 @@
+tbl_quote_name <- function(name) {
+  x <- gsub("`", "``", name, fixed = TRUE)
```
kevinushey: Just double checking -- the intention here is to quote a table name e.g.

    foo`bar

as

    `foo``bar`

?

R/tables_spark.R
```diff
@@ -0,0 +1,73 @@
...
+#'
+#' @param sc A \code{spark_connection}.
+#' @param name The table name.
+#' @param force Force the data to be loaded into memory? This is accomplished
+#'   by calling the \code{count} API on the associated Spark DataFrame.
+#'
+#' @export
+tbl_cache <- function(sc, name, force = TRUE) {
+  countColumns <- function(sc, name) {
+    sql <- paste("SELECT * FROM ", tbl_quote_name(name))
```
kevinushey: This seems like it would be a potentially expensive query if all we need is the number of columns (or is Spark SQL smart enough to not materialize this expression?)

R/data_interface.R
```diff
@@ -307,7 +307,17 @@ spark_data_write_generic <- function(df, path, fileMethod, mode = NULL, csvOptio
   options <- invoke(df, "write")
 
   if (!is.null(mode)) {
-    options <- invoke(options, "mode", mode)
```
javierluraschi: I like that.

R/sdf_saveload.R
javierluraschi: Nice, good to know! Changing.

R/tables_spark.R
javierluraschi: Yes, tried this with this name and this worked as expected. Borrowed this definition from DBI, I believe.

R/tables_spark.R
javierluraschi: Spark is lazy and I remember debugging through similar code that was running correctly. That said, it wouldn't hurt to add "LIMIT = 0" here.

PR #337, opened 2016-11-26
Bugfix: Allow column naming when HEADER=FALSE and columns defined

mnatter:
When calling `spark_read_csv()` with column names defined in the 'columns' named list and the condition HEADER = FALSE, the column names parameter is currently ignored and 'normalized' column names are generated instead of using the supplied names.

Adding the above condition to the `if` statement for normalizing column names fixes the issue.


javierluraschi:
Looks good, merging.

PR #344, opened 2016-11-29
Use get directly rather than duplicating the function

jimhester:
Should make maintainability easier if the definition of `to_host()` ever changes in the future.

See also [`devtools:::"%:::%"`](https://github.com/hadley/devtools/blob/e6df6cf8f410371241a21f32d1960771b4acd7c9/R/utils.r#L15-L17) for a nice syntax if you end up needing to do this more frequently.


javierluraschi:
Thank you!

Issue #308, opened 2016-11-04
Investigate performance regression from 0.4 in ml_linear_regression


javierluraschi:
Feedback from mail: 0.4.21 appears much slower than the official version 0.4 on cran. For example, I fitted a model with ml_linear_regression() using a table of approx. 3 Mio. rows and 150 columns. On my macbook this took about 3 minutes with 0.4 and more than 6 minutes with 0.4.21.

If `do(mod = ml_linear_regression())` takes more than 30 minutes, I receive error message

```
Error in invoke_method.spark_shell_connection(spark_connection(jobj),  : 
  No status is returned. Spark R backend might have failed.
```

kevinushey:
My best guess is that this was caused by a change that forces eager caching of tables when generating a unique ID for each row in a Spark DataFrame, and that caching is slow.

In other words, I wouldn't be surprised if this is the culprit: https://github.com/rstudio/sparklyr/commit/591fd31d3329d76a4a85d6f73050c7a8f763440f

kevinushey:
I wonder if `profvis()` could give us some useful output here, as well? Presumedly not all of the time is spent in the actual model fitting call...

javierluraschi:
@kevinushey the profiler won't help us here since it discards IO time, opened a feature request in `profvis` to investigate if it's feasible to add support for this in `Rprof` and `R` core.

Here are a couple snapshots from Spark with 3M rows using:

```{r}
library(sparklyr)
library(dplyr)
library(nycflights13)

packageVersion("sparklyr")

flights <- nycflights13::flights

sc <- spark_connect(master = "local")

flights10 <- flights[rep(seq_len(nrow(flights)), each=10),]
flights10 <- na.omit(flights10)

flights10_tbl <- copy_to(sc, flights10)

partitions <- flights10_tbl %>% sdf_partition(training = 0.5, test = 0.5, seed = 1099)

system.time({
  partitions$training %>%
    ml_linear_regression(
      response = "dep_delay",
      features = c("arr_time", "sched_arr_time", "arr_delay"))
})

spark_disconnect_all()
```

With `sparklyr` `0.4`:

```
user  system elapsed 
  0.073   0.017   7.412 
```

<img width="1410" alt="screen shot 2016-11-08 at 12 00 02 pm" src="https://cloud.githubusercontent.com/assets/3478847/20114956/490fbe14-a5ab-11e6-896e-c8d24ddb6369.png">

With `sparklyr` `0.4.21`:

```
[...]

PR #345, opened 2016-11-29
Improve performance for dplyr chains with ml

javierluraschi:
Perf regression investigation for https://github.com/rstudio/sparklyr/issues/308, in summary, 4728843 and 591fd31 added valuable functionality to match behavior of `na.action` and to ensure columns are stable across computations; therefore, the computations are needed. Users can avoid the `na.action` functionality by setting `na.action = NULL` in their `ml.options` the other one is required for correct computations.

This PR does improve the case where `na.action` has no side-effect or when it does have a side effect, it caches the output.

The following are perf results for a real `ml_linear_regression` provided by a user over a 260MB partitioned over 200 csv files ran in EMR with a single `m3.xlarge` instance and the data loaded in a S3 bucket.

With `0.4.0` from CRAN:

```
   user  system elapsed 
  1.400   0.144  246.99
```

With master branch this hangs for more than an hour in my particular configuration in EMR.

With master and `na.action = NULL`

```
   user  system elapsed 
  1.936   0.308 373.917 
```

With this PR

```
   user  system elapsed 
  2.060   0.388 478.441 
```

With this PR and `na.action = NULL`

```
   user  system elapsed 
  1.988   0.232 395.106 
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
-Version: 0.4.34
+Version: 0.4.35
 Authors@R: c(
     person("Javier", "Luraschi", email = "javier@rstudio.com", role = c("aut", "cre")),
     person("Kevin", "Ushey", role = "aut", email = "kevin@rstudio.com"),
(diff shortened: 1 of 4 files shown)


javierluraschi:
@kevinushey PR is very scoped at this point since we understand that the impact from new features and fixes is required. The only meaningful fix here is to a) Avoid using the `dropped` `df` if we don't need to drop rows or b) Cache the `dropped` `df` is we do use it. This is to mitigate a hang I experienced in `EMR` given the fact that the dropping was not persisted and therefore, being recalculated over and over again, I would expect Spark to optimize this but it wasn't the case.

Issue #351, opened 2016-12-01
sqlInterpolate() and sqlParseVariables(): Change of interface


krlmlr:
The sparklyr package seems to be the only package that actually implements these generics. I'd like to get rid of the nonstandard argument names and use `conn, sql, ...` as interface (https://github.com/rstats-db/DBI/issues/147), and add an optional `.dots` or `.values` argument (https://github.com/rstats-db/DBI/issues/140#issuecomment-251189284).

Do you plan a CRAN release anytime soon, so that we can coordinate the changes needed in both packages?

javierluraschi:
@krlmlr we are planning to push this to CRAN in about 2 weeks. If you want these changes before that, let me know to figure them out and push them in.

krlmlr:
Thanks. I just released DBI 0.5-13 to GitHub, it contains the breaking changes (rstats-db/DBI#147). Could you make sparklyr work with both CRAN and GitHub versions?

javierluraschi:
@krlmlr The CRAN version for sparklyr 0.4 is already released, so I won't be able to make a change to the current version and sparklyr would be broken for DBI 0.5-13 onwards. I could release a new version of sparklyr 0.5 that bumps the required version of DBI to 0.5.13 but not older ones since the signature changed. When are you planning to release DBI 0.5.13 to CRAN?

Could you shoot me an email to javier at rstudio? I'm buried in github notifications so it's going to be easier to coordinate this by mail. Thanks!

javierluraschi:
From Kirill: "Please prepare for updating the signatures of sqlInterpolate() and sqlParseVariables(), DBI 0.6 will support both but DBI 0.7"

javierluraschi:
Taking care of this in `0.5.3` patch release: https://github.com/rstudio/sparklyr/commits/bugfix/hotfix-0.5.3

krlmlr:
You could add `DBI (>= 0.5-19)` to your `DESCRIPTION`.

javierluraschi:
We took care of this in `sparklyr 0.5.3`.

PR #371, opened 2016-12-07
WIP: fix dbi 0.5.13 breaking change from #351

javierluraschi:
Breaking change coming in DBI, see: https://github.com/rstudio/sparklyr/issues/351

CC: @jjallaire @kevinushey Is there a more generic way to handle this than to embrace the new signature and bump the required version?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,7 @@ Imports:
     methods,
     lazyeval (>= 0.2.0),
     dplyr (>= 0.5.0),
-    DBI (>= 0.4.1),
+    DBI (>= 0.5.13),
     readr (>= 0.2.0),
     digest,
     config,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,7 @@ Imports:
     methods,
     lazyeval (>= 0.2.0),
     dplyr (>= 0.5.0),
-    DBI (>= 0.5.13),
+    DBI (>= 0.4.1),
     readr (>= 0.2.0),
     digest,
     config,
--- a/R/dbi_spark_connection.R
+++ b/R/dbi_spark_connection.R
@@ -20,14 +20,14 @@ setMethod("dbDataType", "spark_connection", function(dbObj, obj) {
   get_data_type(obj)
 })
 
-setMethod("sqlParseVariables", "spark_connection", function(con, sql, ...) {
+setMethod("sqlParseVariables", "spark_connection", function(conn, sql, ...) {
   method <- getMethod("sqlParseVariables", "DBIConnection")
-  method(con, sql, ...)
+  method(conn, sql, ...)
 })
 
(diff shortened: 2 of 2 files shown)


jjallaire:
Is the updated version of DBI on CRAN? If so then I think this is indeed the right change.

javierluraschi:
@jjallaire the DBI change is not yet in CRAN. I'm coordinating when this will happen with @krlmlr to push push a fix on time.

jjallaire:
Okay, sounds good!

krlmlr:
I wonder why you need to specialize if you're only forwarding to the DBI implementation.

javierluraschi:
@krlmlr We do need to specialize, I think, since we morph the `spark_connection` object as a `DBI` connection object, so when you connect to Spark using `sparklyr`, you get a connection that can be used with different packages (dplyr, sparklyr, DBI, etc.).

I just pushed one more iteration to this branch, since we don't use the specialized versions, I'm using the right `setMethod` based on the DBI version. Works for both versions of DBI; however, it does seem a bit hacky. Is this fix a strech? @jjallaire

kevinushey:
@javierluraschi I'm not sure if your most recent change will work as intended. This is code that will be run at package build time; therefore, it will depend on what version of `DBI` is available when `sparklyr` is built.

If a user downloads a `sparklyr` binary from CRAN, but has an older version of `DBI` installed locally, this might break. If you _really_ want to try being clever here, you could move the associated code into the `.onLoad()` function, but I'm not sure if mucking around with S4 method registration at package load time will work as expected, either.

All in all I would vote for not trying to do anything too clever and just synchronize the release with `DBI` (avoid a package version check altogether here).

javierluraschi:
@kevinushey thanks for taking a look here! Some questions... when a package gets built, it replaces the code with an `rdb` file. Not sure how to decompile this binary representation, but I'm curious as to what happens to the `setMethod` directives once "compiled". Does the `rdb` file actually contains a binary format or is it rather still text sources compressed? There is no VM for R so the "compilation" step must be pretty basic, isn't it? Also, how can we validate that this last fix breaks something? I tried compiling from the the IDE the package and then testing against both DBI versions,  but I wasn't to spot any problems while using `soparklyr` with `DBI` operations.

That said, since are skeptic of the last fix, this is good enough reason to revert to the less generic one. However, it would be nice to come up with a different workaround that supports both versions. Otherwise, older versions of sparklyr will be broken with newer versions of DBI and we will have to coordinate another release of sparklyr with the DBI release for this change which would be good to avoid if possible.

I'll leave this PR open since there is no rush of merging this until `DBI` releases again to CRAN at some point.

javierluraschi:
@krlmlr provided a different fix in `DBI` that is backwards compatible, closing this PR.

See comments in: https://github.com/rstats-db/DBI/issues/147

krlmlr:
This still needs to be changed after CRAN release of DBI, but then you can simply use the new signature.

PR #380, opened 2016-12-12
Implement SparklyR side of new package connection contract

jmcphers:
This change implements the SparklyR end of the new RStudio connection contract. Major changes:

- Functions performing actions now generally perform them directly (formerly they sent the code to the IDE, where it was evaluated directly; we can always change this later by having the package use the rstudioapi to emit code to the console)

- The "finder" function is no longer visible to the IDE but used by the package internally 

- Actions specific to Spark (e.g. "show Spark UI") are now implemented in the package instead of in the IDE

Note that this is only the minimal change required for feature parity (e.g. it does not include icons). 

@javierluraschi, this is written in such a way as to compatible with RStudio 1.0 as well as 1.1, but let me know if you'd rather this go on your development branch for the new connection UI.


javierluraschi:
Looks goo to me! However, `sparklyr` is frozen until we release `0.5` to CRAN. We can merge this sometime next week.

jmcphers:
Roger that, just want to be sure you're unblocked and that we don't diverge on changes!

PR #408, opened 2017-01-08
Implement sample_n and sample_frac using tablesample

javierluraschi:
Implement sample_n and sample_frac using tablesample

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
-Version: 0.5.2
+Version: 0.5.3
 Authors@R: c(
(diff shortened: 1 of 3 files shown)


javierluraschi:
Merging this change request from @edgararuiz, his perf results for future reference:

>  I’ve ran a test with a data set that has 11M records. Here are the results:
> 
> | n | function_test         | seconds | rows | columns|
> | --| ------------------ | ----- | ---- | -- |
> |1|cran_sample_n          |19.872|1100|20|
> |2|cran_sample_frac     |17.394|1113 |20|
> |3|cran_sdf_sample      |1.262  |1098|19|
> |4|branch_sample_n      |0.172 |1100 |19|
> |5|branch_sample_frac | 1.180 |1143 |19|
> 
> So, it looks like N sampling is about 20X faster, and Frac about 15X faster than the CRAN version.

Issue #383, opened 2016-12-13
"Failed to execute Livy statement with state waiting" error happens occasionally


fereshtehRS:
Using sparklyr 0.4.41

I have seen this error happening a few times, but not consistently:

```
 fit_sparklyr <- by_cyl %>% do(mod = ml_linear_regression(mpg ~ disp, data = .))
Error in livy_post_statement(sc, statement$code) : 
  Failed to execute Livy statement with state waiting
```

```
sdf_copy_to(sc, iris)
Error in livy_post_statement(sc, statement$code) : 
  Failed to execute Livy statement with state waiting
```

Seems like when the operation takes long we might run into this.

ybjoony:
+1 for this.
Using sparklyr 0.51, this is blocking me from connecting R studio with livy at all as it keeps happening on livyconnection. I can connect to spark from R Studio just fine, and this failed executions start livy sessions normally, but just can't connect it with R Studio.

ybjoony:
created PR for this. [https://github.com/rstudio/sparklyr/pull/421](https://github.com/rstudio/sparklyr/pull/421) any feedback will be appreciated.

javierluraschi:
Thank you @ybjoony

PR #421, opened 2017-01-11
added waiting state

ybjoony:
To fix the issue addressed at https://github.com/rstudio/sparklyr/issues/383, added waiting state to livy connection. As far as I understand, livy's waiting state just means that the statement is waiting for server to execute, as you can see from here: https://github.com/cloudera/livy/commit/b4642b193956b287bb1c1b32bafe4c5a15e71eac

So I thought it would make sense to wait for the waiting status too, since that will eventually transition into running and get finished. Not sure about writing test cases, but after I fixed this, I could make livy work in my cluster. Any feedback will be appreciated.


javierluraschi:
Looks good @ybjoony, thanks for contributing back, merging. Regarding test cases, we currently don't have a template to follow for `Livy` functionality, but we will start add them over time.

PR #440, opened 2017-01-20
Implement ft_count_vectorizer to enable easy use of ml_lda

javierluraschi:
This enable using `ml_lda` with the following example, it was still possible to use `ml_lda` without this change but it would require heavy use of `invoke`:

```
library(janeaustenr)
library(sparklyr)
library(dplyr)

austen_books <- austen_books()
books_tbl <- sdf_copy_to(sc, austen_books, overwrite = TRUE)
first_tbl <- books_tbl %>% filter(nchar(text) > 0) %>% head(2)

first_tbl %>%
  ft_tokenizer("text", "tokens") %>%
  ft_count_vectorizer("tokens", "features") %>%
  ml_lda("features", k = 4)
```

with output:

```
* No rows dropped by 'na.omit' call
An LDA model fit on 1 features

Topics Matrix:
          [,1]      [,2]      [,3]      [,4]
[1,] 0.8996952 0.8569472 1.1249431 0.9366354
[2,] 0.9815787 1.1721218 1.0795771 0.9990090
[3,] 1.1738678 0.8600233 0.9864862 0.9573433
[4,] 0.8951603 1.0730703 0.9562389 0.8899160
[5,] 1.0528748 1.0291708 1.0699833 0.8731401
[6,] 1.1857015 1.0441299 1.0987713 1.1247573

Estimated Document Concentration:
[1] 13.52071 13.52172 13.52060 13.51963
```


R/ml_feature_transformation.R
```diff
@@ -391,3 +391,30 @@ ft_regex_tokenizer <- function(x,
...
+#'
+#' Extracts a vocabulary from document collections.
+#'
+#' @template roxlate-ml-transformation
+#'
+#' @export
+ft_count_vectorizer <- function(x,
+                                input.col = NULL,
+                                output.col = NULL,
+                                min_df = NULL,
```
kevinushey: Can we use `min.df` rather than `min_df` here? I've adopted a convention in the various ML functions that parameter names use `.` as a separator rather than `_`.

javierluraschi: Yes for sure; however, it does make me want to consider renaming them all to `_` at some point since we've named all our public parameters with `_`.

Issue #144, opened 2016-08-03
Dealing with multiple databases


ElianoMarques:
currently for you to leverage the tbl function you have to specify upfront the database. 

Its normal in projects that you need to use tables from multiple databases. 

So would be great if we can find a way to as part of the tbl statement add the database name prior to table name.

```
tbl(sc, "databasename.tablename") 
```

perhaps another feature that could be added is src_databases(sc) similar to src_tbls(sc) 

This will smooth the joins of tables from multiple databases and avoid multiple dbGetQuery(sc, "use database"). 

Thanks

chezou:
I found it depends on dplyr behavior. dplyr assumes the second argument of `tbl()` is only for the table name and it doesn't handle database name.

https://github.com/hadley/dplyr/blob/master/R/sql-escape.r#L241-L248

ElianoMarques:
Great stuff thanks

PR #481, opened 2017-02-10
Handle multiple databases with `tbl()`

chezou:
This PR fixes https://github.com/rstudio/sparklyr/issues/144

I guess there is more appropriate way to handle `sql_escape_ident()`, but it seems to be used for escaping table names, so I use `tbl_quote_name()` instead of `sql_quote()`. If you have better ideas, feel free to comment on it.

I also add `src_dtabases()`, which is similar function as `src_tbls()` to get database list.


javierluraschi:
@chezou looks good, thanks!

javierluraschi:
@chezou had to partially revert this PR with https://github.com/rstudio/sparklyr/pull/500.

Run this code with/without the fix:

```
iris_tbl <- copy_to(sc, iris)
iris_tbl %>% mutate(x = Species)
```

and notice that columns are missing. I don't have a clear suggestion on how to improve this PR, but if you mail me with details, I can try to suggest something sensible. I'm on javier @ rstudio.com.

chezou:
OK. I didn't aware that point. I'll try to dig it.

Issue #496, opened 2017-02-16
temporary spark_serialize_*.csv files don't get deleted


adder:
Dear.
I'm using sparklyr to read hunderds of txt files process in R, copy the resulting dataframe into spark memory and write them to a parquet file.
Sparklyr creates this intermediate spark_serialize_*.csv files in the  /tmp folder in root.
However these files don't get deleted after the data is loaded into file.
Ofter 100's of files these add up to gigabytes and I run out of disk space in my root partition.
I'm not sure if it's expected behavior to keep this files in tmp or not?

I'm under the impression that these files are only created as an intermediate before loading an R data frame in to spark and that they are not needed afterwards.

Best

edgararuiz:
Do the files remain after you close the Spark session (spark_disconnect)?

adder:
Thanks for the fast response,

Yes, the files remain after closing the spark connection.

edgararuiz:
ok, I'd suggest changing the path to the scratch folder to see this has to do with server access: 

**spark.local.dir**	 /tmp	Directory to use for "scratch" space in Spark, including map output files and RDDs that get stored on disk. This should be on a fast, local disk in your system. It can also be a comma-separated list of multiple directories on different disks. NOTE: In Spark 1.0 and later this will be overridden by SPARK_LOCAL_DIRS (Standalone, Mesos) or LOCAL_DIRS (YARN) environment variables set by the cluster manager. - http://spark.apache.org/docs/latest/configuration.html

adder:
Hey,
I did the following:

```
conf = spark_config()
conf$spark.local.dir = paste0(getwd(),'/tmp')
sc = spark_connect(master = "local", version = "2.0.2", hadoop_version = "2.7",config = conf)
```

But this did not help. The files written in this tmp folder by spark get indeed deleted when the spark connection is closed.
The `spark_serialize_*.csv` files however, are written in another tmp folder. I believe it's the R tmp folder (`/tmp/RtmpkkztvF`).
These files remain upon closing the spark connection.

adder:
Great! Thanks a lot!

javierluraschi:
Haven't given this much thought, but worth exploring cleaning temp files more aggressively than waiting until the r session restarts.

yl790:
[...]

PR #498, opened 2017-02-17
WIP: Immediately remove serialized temp csv files

javierluraschi:
See https://github.com/rstudio/sparklyr/issues/496. Worth considering an alternate fix @kevinushey or looking good to you?


javierluraschi:
We really can't take this PR as it is since some operations are lazy...

PR #501, opened 2017-02-18
Handling database name again

chezou:
This PR is retry of https://github.com/rstudio/sparklyr/pull/481 fixing https://github.com/rstudio/sparklyr/pull/500

I find out `sql_escape_ident()` handles array like variables, such as column names, and `tbl_quote_name()` can't handle array because it uses `strsplit()`. So, I add the heuristics if the length of an object is 1, it might be a table name.

### DIFF
--- a/R/dplyr_spark_connection.R
+++ b/R/dplyr_spark_connection.R
@@ -1,6 +1,6 @@
 #' @export
 sql_escape_ident.spark_connection <- function(con, x) {
-  sql_quote(x, '`')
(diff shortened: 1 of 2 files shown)


chezou:
I guess dplyr has a responsibility to handle database name essentially, but I think Spark uses the table join rather than RDBs.

javierluraschi:
Thanks for the update @chezou, merging.

chezou:
Thanks!

PR #516, opened 2017-02-23
Update spark_compile.R

AkhilNairAmey:
`scala_path` pointed at non-existant `java` folder in package root resulting in no scala files being pulled into the jar

### DIFF
--- a/R/spark_compile.R
+++ b/R/spark_compile.R
@@ -42,7 +42,7 @@ spark_compile <- function(jar_name,
   java_path <- file.path(root, "inst/java")
   jar_path <- file.path(java_path, jar_name)
 
-  scala_path <- file.path(root, "java")
+  scala_path <- file.path(root, "inst/scala")
   scala_files <- list.files(scala_path, pattern = "scala$", full.names = TRUE)
 
   # apply user filter to scala files


kevinushey:
Thanks for catching that!

AkhilNairAmey:
@kevinushey No worries - however I did realise soon after submitting that, that reason it was like that is presumably because you compile the `sparklyr` scala files with `sparklyr::spark_compile`, for which that was the correct folder path. 

Presumably that commit would mess up your workflow - would it be safe to simply move the `java` folder to `inst/scala` for consistency?  Seems like that's where they're supposed to be

javierluraschi:
@AkhilNairAmey had to revert this one since it breaks a couple dependencies like the http://github.com/javierluraschi/sparkhello package that some people are using as a base to develop custom packages.

Could you provide an example of code that requires this change?

My guess is that you were expecting that `compile_package_jars()` would do something meaningful in `sparklyr`? If that's the case, `compile_package_jars()` doesn't do anything meaningful unless you are running this function from within the `sparklyr` sources, in which case, it would compile the internal `jars`. Otherwise, this function is only used by package creators from their own custom extension.

Do let me know if you think I'm missing something.

AkhilNairAmey:
This was my mistake - I've also been using `sparkhello` to write an extension, however I was somehow using an old version with the old file structure and `README` (I think I must've been on the jjallaire fork). 

The file structure has since been updated to move the `inst/scala` to `java` presumably to comply with CRAN guidelines.

Sorry about that :(

javierluraschi:
@AkhilNairAmey no worries, we all make mistakes, thanks for sending this our way!

PR #521, opened 2017-02-27
Support for dplyr 0.6

javierluraschi:
Want to write some `dplyr` unit tests before we merge this, so opening this one as WIP. Changes:

 1. `dplyr` now supports `LIMIT` in queries, which allows us to remove our custom code.
 2. `dplyr` not longer requires `USING` which was not supported in `Spark 1.6` and therefore, custom code can be removed.
 3. `mutate` no longer compatible due to `lazyeval` changes, `0.6` will require this change and therefore, this has been backported to this branch https://github.com/rstudio/sparklyr/commits/bugfix/hotfix-0.5.3 to patch CRAN after `dplyr` releases.
 4. I believe we no longe need `sql_build.tbl_spark`, especially since it gets in the way of triggering `dplyr`'s new query optimizer.


kevinushey:
Are these changes compatible with `dplyr 0.5.0` or are we waiting for the next `dplyr` release to merge this?

javierluraschi:
@kevinushey they are "compilable"; but they do break a couple things:
 1. **Joins**: They won't work in `dplyr 0.5` and `Spark 1.6` since I'm removing the workaround for not using `USING`
 2. **Limit**: This will probably break commands that require this function: `head`, etc.

My plan is to bump the `dplyr` version to `5.0.9000` and force anyone using `sparklyr` latest to also use `dplyr` latest for about 3 weeks until `dplyr` releases to CRAN in which case we can bump again the required version to `6.0`.

BTW. There is a breaking change that we do need in `sparklyr`, I have that change isolated in this branch that we would push to CRAN after `dplyr` releases: https://github.com/rstudio/sparklyr/commits/bugfix/hotfix-0.5.3

javierluraschi:
@hadley our own `mutate` was added since, at the time, I was trying to solve this:

```
iris %>% mutate(a = Petal.Width, b = a)
```

which works as expected for data frames, but not for `SQL` backends. In retrospect, I should have sent a PR for this to `dplyr`, do you want me to send one? I would scope this optimization to only `mutate` calls that self reference new columns.

The commit that introduced this was: https://github.com/rstudio/sparklyr/commit/b225cf49b56bdb05ea55ff79c65aa2a145395733

javierluraschi:
This one is good to go now; however, `sample_n` and `sample_frac` require this fix: https://github.com/hadley/dplyr/pull/2480

hadley:
Yeah, could you please send a PR? (Or just an issue and I can figure out how to solve it)

javierluraschi:
Opened: https://github.com/hadley/dplyr/issues/2481

DESCRIPTION
```diff
@@ -39,6 +38,8 @@ Imports:
     base64enc,
     shiny,
     rstudioapi
+Remotes:
+    dplyr (>= 0.5.0.9000)
```
kevinushey: You want this to be e.g. `hadley/dplyr`; that signals that `dplyr` should be installed using `devtools::install_github("hadley/dplyr")`.

javierluraschi: Thanks, we get now a `Namespace dependency not required` in `R CMD check`, but If I remove the `@import dplyr` refs we get an error referencing `collect` from `dplyr` 😕

hadley: You need it in both remotes and imports - remotes is just a way to tell devtools where to get it from. R doesn't know anything about it.

javierluraschi: Thank you @hadley!

javierluraschi:
@kevinushey merging this one, I was missing `dplyr` in the `imports` as Hadley pointed out.

PR #535, opened 2017-03-07
fix NPE when passing null to a Seq parameter via invoke

farrellm:
Using invoke to call a Scala function with a Seq parameter with a null argument results in an NPE. This change fixes the bug by adding a couple null checks. Additionally, I added some unit tests for invoke_static that exercise the bug.

I wasn't sure the procedure for updating the Scala code, so I ran "configure.R" and included the resulting artifacts (inst/livy/invoke.scala & inst/java/sparklyr-*.jar). If this is not necessary, let me know and I would be happy to remove them from the PR.


kevinushey:
LGTM; thanks! If you haven't already, can you provide us with a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf)?

javierluraschi:
LGTM as well.

farrellm:
No, I have not completed the CLA, but happy to. However, the email in the CLA (jj AT rstudio.org) bounced. Is there a new address?

kevinushey:
Sorry; I was under the impression that CLAs would be required for `sparklyr` but they are not, so don't worry!

PR #552, opened 2017-03-14
fix http response validation for livy

kevinykuo:
From Windows desktop I was getting 
```
> sc <- spark_connect(master = "foobar", method = "livy")
Error in if (all.equal(status_code(req), 401)) { : 
  argument is not interpretable as logical
```


javierluraschi:
Thanks @kevinykuo! Merging.

Issue #313, opened 2016-11-07
sparklyr als getting a java.lang.NullPointerException error when predicting


blacknred0:
Hi - 
Not sure if this is an issue from my side or not.  I am currently trying to use ALS Matrix Factorization [http://spark.rstudio.com/reference/sparklyr/latest/ml_als_factorization.html](url) in sparklyr (v0.4) with R 3.3.2 (Sincere Pumpkin Patch) on *nix (Mac and Linux).  I am also using Apache Spark compile version 2.0.1 with hadoop 2.7.

I found this issue [https://github.com/rstudio/sparklyr/issues/139](url) which it was fixed, but it doesn't seem to work on my particular case, not sure why.  The file that I am using does not have null/NA/blanks values (since I am using spark-2.0.1-bin-hadoop2.7/data/mllib/als/test.data)

Here is the code that I am using

``` R
library(sparklyr)
library(dplyr)

sc <- spark_connect(master = "local", config = spark_config())
test <- read.table('~/Applications/spark-2.0.1-bin-hadoop2.7/data/mllib/als/test.data', 
   header = FALSE, sep = ',')

test_tbl <- copy_to(sc, test, overwrite = TRUE)
als_model <- test_tbl %>%
  ml_als_factorization(user.column = "V1", item.column = "V2", rating.column = "V3", 
  rank = 10L, regularization.parameter = 0.01, iter.max = 10L, ml.options = ml_options())
predictions = als_model$.model %>% invoke("transform", spark_dataframe(test_tbl))
#sdf_predict(als_model, test_tbl)
sdf_predict(predictions, test_tbl) %>%
  collect
spark_disconnect(sc)
```

error
`Error: java.lang.NullPointerException
	at scala.collection.immutable.StringOps$.length$extension(StringOps.scala:47)
	at scala.collection.immutable.StringOps.length(StringOps.scala:47)
	at scala.collection.IndexedSeqOptimized$class.isEmpty(IndexedSeqOptimized.scala:27)
	at scala.collection.immutable.StringOps.isEmpty(StringOps.scala:29)
	at scala.collection.TraversableOnce$class.nonEmpty(TraversableOnce.scala:111)
	at scala.collection.immutable.StringOps.nonEmpty(StringOps.scala:29)
	at org.apache.spark.ml.attribute.AttributeGroup.<init>(AttributeGroup.scala:42)
	at org.apache.spark.ml.attribute.AttributeGroup.<init>(AttributeGroup.scala:65)
[...]

PR #565, opened 2017-03-20
Add `ml_als_factorization()` options

### LINKED ISSUE #561: Implicit Function calls for ALS
I'm using the ALS algorithm and the model doesn't have a parameter for using Implicit ALS. Here's the scala code that I would normally use for the same purpose. 

val als = new ALS()
  .setMaxIter(5)
  .setRegParam(0.01)
  .setImplicitPrefs(true) // Not available
  .setUserCol("userId")
  .setItemCol("movieId")
  .setRatingCol("rating")

chezou:
This PR adds `ml_als_factorization()` options as follows:
- Implicit matrix factorization. close #561 
- Nonnegative least square option

I couldn't find a prediction example of ALS and finally found #313, I created a simple example of patched version. http://rpubs.com/chezou/sparklyr-als

### DIFF
--- a/R/ml_alternating_least_squares.R
+++ b/R/ml_alternating_least_squares.R
@@ -8,6 +8,8 @@
 #' @param item.column The name of the column containing item IDs.
 #' @param rank Rank of the factorization.
 #' @param regularization.parameter The regularization parameter.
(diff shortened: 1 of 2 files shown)


kevinushey:
LGTM -- thanks!

PR #567, opened 2017-03-20
Support saving to default hive path for spark_write_table

javierluraschi:
Intended use:

```
spark_write_table(iris_tbl, "iris_named")
iris_named_tbl <- spark_read_table(sc, "iris_named")
```

@edgararuiz found out this is how customers expect to use read/write tables into Hive which avoids having to specify the `path`.

### DIFF
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -394,12 +394,13 @@ spark_load_table <- function(sc,
 #' Writes a Spark DataFrame into a Spark table.
 #'
 #' @inheritParams spark_write_csv
+#' @param name The name to assign to the newly generated table.
 #' @param mode Specifies the behavior when data or table already exists.
 #'
 #' @family Spark serialization routines
 #'
 #' @export
-spark_write_table <- function(x, name = NULL, mode = NULL, options = list()) {
+spark_write_table <- function(x, name, mode = NULL, options = list()) {
   UseMethod("spark_write_table")
 }
 
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -426,7 +426,7 @@ spark_write_table.tbl_spark <- function(x, name, mode = NULL, options = list()) 
 
   if (spark_version(sc) < "2.0.0" && spark_master_is_local(sc$master)) {
     stop(
-      "spark_write_table is not supported in local clusters for ",
+      "spark_write_table is not supported in local clusters for Spark ",
       spark_version(sc), ". ",
       "Upgrade to Spark 2.X or use this function in a non-local Spark cluster.")
   }
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -424,6 +424,13 @@ spark_write_table.tbl_spark <- function(x, name, mode = NULL, options = list()) 
   sqlResult <- spark_sqlresult_from_dplyr(x)
   sc <- spark_connection(x)
 
+  if (spark_version(sc) < "2.0.0" && spark_master_is_local(sc$master)) {
+    stop(
+      "spark_write_table is not supported in local clusters for ",
+      spark_version(sc), ". ",
+      "Upgrade to Spark 2.X or use this function in a non-local Spark cluster.")
+  }
+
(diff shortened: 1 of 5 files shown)


R/data_interface.R
```diff
@@ -350,16 +350,14 @@ spark_data_write_generic <- function(df, path, fileMethod, mode = NULL, csvOptio
...
                              options = list(),
                              repartition = 0,
                              memory = TRUE,
                              overwrite = TRUE) {
 
   if (overwrite) spark_remove_table_if_exists(sc, name)
 
-  operation <- if (spark_version(sc) < "2.0.0") "load" else "table"
-  df <- spark_data_read_generic(sc, path, operation, options)
+  df <- spark_data_read_generic(sc, name, "table", options)
```
kevinushey: Does this work with Spark 1.6? (is the indirection above now unneeded?)

R/data_interface.R
```diff
@@ -422,21 +419,19 @@ spark_save_table <- function(x, path, mode = NULL, options = list()) {
...
 
 #' @export
-spark_write_table.spark_jobj <- function(x, path, mode = NULL, options = list()) {
+spark_write_table.spark_jobj <- function(x, name, mode = NULL, options = list()) {
   spark_expect_jobj_class(x, "org.apache.spark.sql.DataFrame")
   sc <- spark_connection(x)
 
-  operation <- if (spark_version(sc) < "2.0.0") "save" else "saveAsTable"
-  spark_data_write_generic(x, spark_normalize_path(path), operation, mode, options)
+  spark_data_write_generic(x, name, "saveAsTable", mode, options)
```
kevinushey: Same comment here re: Spark 1.6?

R/data_interface.R
javierluraschi: @kevinushey yes, it works in Spark 1.6 but NOT in Spark 1.6 in local mode. In Spark 1.6 local you get this exception:

```
org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 6.0 failed 1 times, most recent failure: Lost task 1.0 in stage 6.0 (TID 9, localhost): java.io.IOException: Mkdirs failed to create file:/user/hive/warehouse/iris_named/_temporary/0/_temporary/attempt_201703201501_0006_m_000001_0 (exists=false, cwd=file:/Users/javierluraschi/RStudio/sparklyr)
```

The problem is that somehow, Spark 1.6 does not set the right permissions on the Hive folder. I verified that `chmod g+w /user/hive/` solves this issue.

While we could fix this ourselves, I'm a bit hesitant since we currently run without workarounds in OS X (as opposed to several config changes in Windows). I'm thinking there is a set of non-core functions that are not supported in local environment, thoughts?

edgararuiz: Could an check be added to only default to this behavior if the master is 'yarn-client'?

kevinushey: I think it's relatively unlikely that users will want to run older versions of Spark locally, so perhaps this is okay to leave behind.

R/data_interface.R
javierluraschi: This one actually works fine since by default everyone gets `r` permissions in the `hive` folder.

R/data_interface.R
javierluraschi: @edgararuiz Well, we don't have fallback behavior since the previous behavior was to save this with `path=name` which will save the data on the current directory with the given `name` as `path`, but you know what, we can throw an `unsupported` error when running in `master` mode. Let me add that.

javierluraschi: @edgararuiz @kevinushey Alright, now we throw the following for Spark 1.6 and `master="local"`:

```
> spark_write_table(iris_tbl, name = "iris_named")
Error in spark_write_table.tbl_spark(iris_tbl, name = "iris_named") : 
  spark_write_table is not supported in local clusters for 1.6.0. Upgrade to Spark 2.X or use this function in a non-local Spark cluster.
```

Issue #568, opened 2017-03-21
Error in Sparklyr extension creation


mshafieluru:
Hi

I am trying to create an extension as mentioned in the git hub on windows.
I am getting below error.
sparklyr::compile_package_jars() Error: failed to discover scala-2.10 compiler -- search paths were:
"C:/Users/abc/Documents/scala"

I have scala 2.10.6 installed on my D drive in windows and it is in Path. Why it is trying to look in another path ? i tried creating shortcut as well but did not work.

Help me to resolve this issue

javierluraschi:
@mshafieluru Is there a reason why `scala` can't be installed under `Documents`?

The reason why having it in the `scalac` in the `PATH` is not enough is that `sparklyr` requires multiple versions of the `scalac` compiler since `Spark 1.6` and `Spark 2.0` use different toolchains. Therefore, in Windows, the `2.10` and `2.11` compilers need to be installed under `Docuemnts/scala`.

I'm pushing a change to make this a bit easier by introducing a `download_scalac` helper function to  get us to easily install this in Windows (since the msi does not support installing multiple versions) and also to be more flexible on the `scala` versions we support, will push this soon and update this thread. However, it would help to know if you are restricted from using `Documents/scala` as the install directory before I make this change.

javierluraschi:
@mshafieluru using `compile_spark_jars` required a few steps, we've made a few changes to make this easier, could you try the following?

```
devtools::install_github("rstudio/sparklyr")
sparklyr::download_scalac()
sparklyr::compile_package_jars()
```

Those steps still require the `JAVA_HOME` environment variable to be set to, say, "c:\Program Files\java\jdk1.8.0_73".

If you can't upgrade `sparklyr` at the moment, you can still follow the workarounds from this PR to set your environment ready to compile extensions in Windows: https://github.com/rstudio/sparklyr/pull/570

mshafieluru:
Hi 
My suggestion is instead of hard coding the installation directories, kindly look for SPARK_HOME or SCALA_HOME or JAVA_HOME variables and take it from there. Users like me can install java or scala anywhere in system.

you can easily get the version of scala from these variables. depending on the version you can invoke your respective funtions.

Anyhow, you knows it better. 

regards
Shafi

mshafieluru:
devtools::install_github("rstudio/sparklyr")
Error in curl::curl_fetch_disk(url, x$path, handle = handle) : 
  Timeout was reached

[...]

PR #570, opened 2017-03-21
Simplify compilation in Windows

javierluraschi:
See https://github.com/rstudio/sparklyr/issues/568, which would require the following workarounds to get `compile_package_jars()` working:
 1. Download Scala 2.10 as zip, extract to `Documents/scala`, rename to `scala-2.10`
 2. Download Scala 2.11 as zip, extract to `Documents/scala`, rename to `scala-2.11`
 3. Find `JAVA_HOME` and add `bin` to `PATH` environment variable.

Instead, introducing `download_scalac` do perform (1) and (2) and add support to reuse `JAVA_HOME` which is usually already set while installing `JDK`.


javierluraschi:
This needs a bit more work since in Windows the `Scala` installer do not support installing multiple versions.

Issue #411, opened 2017-01-08
Collecting string columns with new line characters splits into multiple rows


javierluraschi:
Hit an issue where collecting a data frame with data `"this\ntest"` would collect as two rows instead of one. This could also break collection if the rows get out of sync, I believe. See for instance `sdf_deserialize_column` which splits based on `\n`, etc.

kevinykuo:
So I've been playing around with this and couldn't figure this out: if we escape `\n` on the JVM side (via `_.replaceAll("\n", "\\n")` in `collectColumnString` it somehow gets unescaped backed to `n` during deserialization. I.e. `aa\\nbb` in JSON gets collected as `aabb` after `spark_read_json`. Thoughts?

kevinushey:
One thought I had before was whether we could use the ASCII record separator (`\30`) but if I recall correctly this did not play well with `livy` / JSON fields.

kevinykuo:
@javierluraschi I've tried a few ways (mostly escaping/unescaping) to work around this to no avail. I think the issue breaks a lot of use cases and should be looked at soon.

kevinushey:
I can't even copy a `data.frame` containing newlines into Spark, so it seems like there's far more fundamental issues than just what's in sparklyr.

```R
library(sparklyr)
sc <- spark_connect(version = "2.1.0", master = "local")
df <- data.frame(
  x = 1:3,
  y = c("a\nb", "c\nd", "e\nf"),
  stringsAsFactors = FALSE
)
copy_to(sc, df)
```

vomits

```
 Error: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 1.0 failed 1 times, most recent failure: Lost task 0.0 in stage 1.0 (TID 1, localhost, executor driver): java.lang.NumberFormatException: For input string: "b""
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)
	at scala.collection.immutable.StringLike$class.toInt(StringLike.scala:272)
	at scala.collection.immutable.StringOps.toInt(StringOps.scala:29)
	at org.apache.spark.sql.execution.datasources.csv.CSVTypeCast$.castTo(CSVInferSchema.scala:250)
	at org.apache.spark.sql.execution.datasources.csv.CSVRelation$$anonfun$csvParser$3.apply(CSVRelation.scala:125)
[...]

PR #574, opened 2017-03-24
Support collecting sets with new lines to fix #378 and #411

### LINKED ISSUE #378: Failure when JSON fields contain `\n` or `\r` characters
Given the following json file:

```
path <- '/tmp/jsonissue.json'
cat('{"name":"Tim", "desc":"Multiline\\nText"}\n{"name":"Yates", "desc":"Single Text"}', file=path)
```

When we attempt to parse this with `sparklyr`:

```
library(sparklyr)
sc <- spark_connect(master = "local", config = spark_config())
stock <- spark_read_json(sc, "stock", path)
head(stock)
spark_disconnect_all()
```

We get the following error:

```
Source:   query [2 x 2]
Database: spark connection master=local[4] app=sparklyr local=TRUE

Error: Variables must be length 1 or 3.
Problem variables: 'name' 
```

Running the same thing in SparkR gives the expected result:

```
library(SparkR)
sc <- sparkR.session(master = 'local')
stock <- read.json(path)
head(stock)
             desc  name
1 Multiline\nText   Tim
2     Single Text Yates
```

The problem is in the way sparklyr is parsing the `\\n` chars in the input json, as can be seen if we just go down to the single record in the json:

```
cat('{"name":"Tim", "desc":"Multiline\\nText"}', file=path)
sc <- spark_connect(master = "local", config = spark_config())
stock <- spark_read_json(sc, "stock", path)
head(stock)
```

Which gives the result:

```
Source:   query [1 x 2]
Database: spark connection master=local[4] app=sparklyr local=TRUE

       desc  name
      <chr> <chr>
1 Multiline   Tim
2      Text   Tim
```

As you can see, the multiline field is applied across multiple records 😿

javierluraschi:
Avoid using `\n` as the line separator to collect data.

See #378 and #411.


kevinushey:
Awesome, thanks for tackling this!

PR #575, opened 2017-03-24
Added set_spark_home_env_var()

richierocks:
It's useful for users to be able to set the `SPARK_HOME` environment variable without having to hunt for where it is.


R/set_spark_home_env_var.R
```diff
@@ -0,0 +1,27 @@
...
+#' 
+#' Set the \code{SPARK_HOME} environment variable. This slightly speeds up some 
+#' operations, including the connection time.
+#' @param path A string containing the path to the installation location of
+#' Spark. If \code{NULL}, the path to the most latest Spark/Hadoop versions is
+#' used. 
+#' @param verbose Logical. Should the function explain what is it doing?
+#' @return The function is mostly invoked for the side-effect of setting the 
+#' \code{SPARK_HOME} environment variable. It also returns \code{TRUE} if the 
+#' environment was successfully set, nad \code{FALSE} otherwise.
```
javierluraschi: Typo: `nad`

R/set_spark_home_env_var.R
```diff
@@ -0,0 +1,27 @@
...
+#' @examples
+#' \dontrun{
+#' # Not run due to side-effects
+#' set_spark_home_env_var()
+#' }
+#' @export
+set_spark_home_env_var <- function(path = NULL, verbose = is.null(path)) {
```
javierluraschi: I think you could change the signature to:

```
function(path = spark_home_dir()) {
})
```

I would also consider removing `verbose` as a parameter and use instead:

```
verbose <- sparklyr_boolean_option("sparklyr.verbose")
```

Which will be `FALSE` by default but users can change this with `options(sparklyr.verbose = TRUE)`. We don't use this option in many places but it's certainly an option we support.

javierluraschi: Nit: Since you are creating a new file, it would make sense to name it `spark_home.R` and potentially also move the `spark_home_dir` function into this file, up to you, otherwise we can move that later.

javierluraschi: I would rather name this function `spark_home_set` since we have a similar one named `spark_home_dir` that will help users discover them with ease.

javierluraschi:
Thank you @richierocks! A few minor comments inline, but otherwise looking good!

R/set_spark_home_env_var.R
richierocks: Fixed via https://github.com/rstudio/sparklyr/pull/575/commits/6b0fe81aae98e169cc6cb5adf3c0257c5c414415

R/set_spark_home_env_var.R
richierocks: Fixed via https://github.com/rstudio/sparklyr/pull/575/commits/6d370e847d802e73dbd4eae8e1b628320665b311

R/set_spark_home_env_var.R
richierocks: I've kept the `verbose` argument, since it's nice for users to be able to control it, but the value now defaults to the `sparklyr.verbose` option. See https://github.com/rstudio/sparklyr/pull/575/commits/95e02ce3078ec8e33e092d6dc72f3e0dc03dd1e8

richierocks: I've moved `spark_home_dir()` to `spark_home.R` via https://github.com/rstudio/sparklyr/pull/575/commits/29f10d109a20f3f744401d4a0e9f5d155a69edda.

javierluraschi:
Nice, thank you @richierocks! Merging...

Issue #583, opened 2017-03-30
Error: object 'db_data_type' not found whilst loading namespace 'sparklyr'


navdeep-G:
I am seeing the following issue when trying to download sparklyr from github:

`Error: object 'db_data_type' not found whilst loading namespace 'sparklyr'`

AlexYoung757:
Error: object 'db_data_type' not found whilst loading namespace 'sparklyr' . i also find this error

GriffinRidgeback:
This is NOT fixed! Tried to install on a new R environment for work I HAVE to do in sparklyr and still got:

Error : object 'db_data_type' not found whilst loading namespace 'sparklyr'
I saw somewhere that this had been fixed but evidently not.

Why is this still not working? I have a customer who is depending on me being able to transform their code into sparklyr format but I have no environment in which to do this.

Please help!! What diagnostic information can I provide that will help in this? Has anyone installed this package successfully? Under what versions of R and installed packages did it work?

Thanks!!

kevinykuo:
@AlexYoung757 @GriffinRidgeback for the time being installing from @navdeep-G's repo worked for me

navdeep-G:
@kevinykuo @AlexYoung757 @GriffinRidgeback Should be fixed now. @javierluraschi merged my PR this morning. Will close this issue.

PR #584, opened 2017-03-30
Fix issue with NAMESPACE not finding certain S3 methods

navdeep-G:
* This PR fixes the following issue: https://github.com/rstudio/sparklyr/issues/583

### DIFF
--- a/R/dplyr_spark.R
+++ b/R/dplyr_spark.R
@@ -30,9 +30,14 @@ src_desc.src_spark <- function(x) {
 }
 
 #' @export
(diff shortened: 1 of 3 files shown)


javierluraschi:
We can't overwrite the base calls (say `sql_build`) since this would prevent using `dplyr` with other backends when `sparklyr` is loaded, but since this `dplyr` change is breaking loading `sparklyr`, I do agree there is urgency to get this fixed; therefore, merging. Thank you @navdeep-G!

PR #587, opened 2017-03-30
Protect against older contracts that don't support listObjectTypes

javierluraschi:
Older connection viewer contracts are breaking `spark_connect` since `listObjectTypes` is not defined. We need to ignore this for a bit while people move out to newer versions of the IDE daily builds.


javierluraschi:
Taking care of this with newer PR.

PR #591, opened 2017-03-31
Supporting sparklyr in Databricks clusters

falaki:
## What changes are proposed in this pull request?
This patch adds support for Databricks Spark clusters by adding a method type to `spark_connect`. When `method = 'databricks'`, there is no need for any additional arguments to `spark_connect`.

## Implementation details
The new internal `databricks_connection()` function, does two main things:
1. Makes an internal Databricks call to start the sparklyr backend on the currently attached cluster. This call passes the current location of sparklyr JARs to Databricks backend. The backend will perform the work on searching and find the right JAR and will call its main method.
2. Uses the gateway port returned from step 1 to establish gateway and backend connections

We think sparklyr is going to be an active project with multiple releases in coming months. As a result, we think hard coding the JARs in Databricks spark images will tie users to older versions of the project. We will continue using the JARs that are installed by the R package.

In Databricks notebooks and jobs, the JVM and Spark Context are managed by Databricks. Therefore, applications need to use the existing context and using `spark-submit` will not work. Also, in Databricks notebooks SparkR connection is already established and provides a mechanism for calling into the JVM. We are using SparkR to bootstrap the sparklyr connection

### DIFF
--- a/R/databricks_connection.R
+++ b/R/databricks_connection.R
@@ -1,0 +1,64 @@
+databricks_connection <- function(config, extensions) {
+  tryCatch({
+    callSparkR <- get("callJStatic", envir = asNamespace("SparkR"))
+    gatewayPort <- as.numeric(callSparkR("com.databricks.backend.daemon.driver.RDriverLocal",
+                                         "startSparklyr",
+                                         # In Databricks notebooks, this variable is in the default namespace
+                                         get("DATABRICKS_GUID", envir = .GlobalEnv),
+                                         # startSparklyr will search & find proper JAR file
+                                         system.file("java/", package = "sparklyr")))
+  }, error = function(err) {
+    stop(paste("Failed to start sparklyr backend:", err$message))
+  })
+
+  tryCatch({
+    gatewayInfo <- spark_connect_gateway(gatewayAddress = "localhost",
+                                        gatewayPort = gatewayPort,
+                                        sessionId = gatewayPort,
+                                        config = config,
+                                        isStarting = TRUE)
+  }, error = function(err) {
+    stop(paste("Failed to open connection to sparklyr gateway:", err$message))
+  })
+
+  output_file <- tempfile(fileext = "_spark.log")
+  error_file <- tempfile(fileext = "_spark.err")
+
+  tryCatch({
+    # set timeout for socket connection
+    timeout <- spark_config_value(config, "sparklyr.backend.timeout", 30 * 24 * 60 * 60)
+    backend <- socketConnection(host = "localhost",
+                                port = gatewayInfo$backendPort,
+                                server = FALSE,
+                                blocking = TRUE,
+                                open = "wb",
+                                timeout = timeout)
+  }, error = function(err) {
+    stop(paste("Failed to open connection to sparklyr backend:", err$message))
+  })
+
+  # create the databricks connection
+  sc <- structure(class = c("spark_connection", "spark_shell_connection"), list(
+    # spark_connection
+    master = "databricks",
+    method = "shell",
+    app_name = "Databricks",
+    config = config,
+    # spark_shell_connection
+    spark_home = Sys.getenv("SPARK_HOME"),
+    backend = backend,
+    monitor = gatewayInfo$gateway,
+    output_file = output_file
+  ))
+
+  # stop shell on R exit
+  reg.finalizer(baseenv(), function(x) {
+    if (connection_is_open(sc)) {
+      stop_shell(sc)
+    }
+  }, onexit = TRUE)
+
+  sc
+}
+
+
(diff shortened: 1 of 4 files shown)


R/spark_shell.R
```diff
@@ -7,6 +7,68 @@ shell_connection_validate_config <- function(config) {
   config
 }
 
+databricks_connection <- function(config, extensions) {
+  tryCatch({
+    gatewayPort <- as.numeric(SparkR:::callJStatic("com.databricks.backend.daemon.driver.RDriverLocal",
```
jjallaire: You need to call the sparklyr equivalent here, see http://spark.rstudio.com/reference/sparklyr/latest/invoke.html

falaki: The problem is that the sparklyr spark connection (sc) is not setup yet.

javierluraschi: Right, need to replace `SparkR:::callJStatic` with `invoke_static`.

R/spark_shell.R
```diff
@@ -7,6 +7,68 @@ shell_connection_validate_config <- function(config) {
   config
 }
 
+databricks_connection <- function(config, extensions) {
```
javierluraschi: I would create a new file for `databricks_connection` named `R/databricks_connection.R`

R/spark_shell.R
```diff
@@ -7,6 +7,68 @@ shell_connection_validate_config <- function(config) {
...
+                                open = "wb",
+                                timeout = timeout)
+  }, error = function(err) {
+    stop(paste("Failed to open connection to sparklyr backend:", err$message))
+  })
+
+  # create the databricks connection
+  sc <- structure(class = c("spark_connection", "spark_shell_connection"), list(
+    # spark_connection
+    master = "databricks",
```
javierluraschi: Would anything break on your side if we change `master="databricks"` into, for instance, `master="remote"`? Or does the infrastructure assumes this needs to be set to `databricks`?

R/spark_shell.R
falaki: What to do about `sc`?

javierluraschi: Actually, this comes before the connection is opened so `invoke_static` won't work here. This call assumes that `SparkR` is configured in their cluster and uses `SparkR` to make the `RDriverLocal` call.

To have this be a clean call, we could either:

1. Always start the `sparklyr` backend in databricks and the port is read from a config.
2. Use some other mechanism to start `sparklyr` in databricks, say, with a call to a REST endpoint.

In the sort term, we would have to mask the call to `SparkR` with something like:

```
if (.rs.isPackageInstalled("SparkR")) {
      callSparkR <- get("callJStatic", envir = asNamespace("SparkR"))
      gatewayPort <- as.numeric(callSparkR(...))
}
```

otherwise, we won't be able to push this to CRAN since `SparkR` is not a known package.

javierluraschi:
@falaki LGTM. Thanks for updating the docs, that should fix travis.

R/databricks_connection.R
```diff
@@ -0,0 +1,64 @@
+databricks_connection <- function(config, extensions) {
+  tryCatch({
+    callSparkR <- get("callJStatic", envir = asNamespace("SparkR"))
```
jjallaire: One way we've managed these sorts of "ambient" dependencies (where we want to call something not on CRAN that we *know* is available) is to have the environment primed with a callback. So for example you could set an R option at Notebook startup like this:

```r
options(sparkr.startSparklyr = function() { # do what's needed to start sparklyr and return the port })
```

Then the client code would do this:

```r
startSparklyr <- getOption("sparkr.startSparklyr")
if (!is.null(startSparklyr))
   gatewayPort <- startSparklyr()
```

This is pretty much identical in execution to what we have but it's a bit more decoupled (the implementation of the function called via the `sparkr.startSparklyr` option can be changed at any time). This is also a bit more kosher with CRAN as using asNamespace may work around their automated checks but it's not really in the spirit of not calling packages that aren't on CRAN. Let me know if this makes sense.

jjallaire: The option name I suggested might be wrong, perhaps:

```r
options(databricks.startSparklyr = function() {})
```

jjallaire: Granted this is not technically an "option" but it's the one shared sink that the R environment provides to store references to functions in a decoupled way (we have used this frequently to allow R packages to call back into RStudio). Javier, if you wanted to make it generic you could establish a standard that the following would be called:

```r
options(sparklyr.remote_gateway = function() {})
```

Then Databricks could set that value in their environment, but other environment that provide other remote gateways could set it to something different (this might dovetail well with your idea of supporting a generic `method = "remote"` option).

PR #610, opened 2017-04-11
Return object type information only for the new connection contract

jmcphers:
This change fixes an issue with the development version of sparklyr and the stable release of RStudio. The development version of RStudio expects `listTables` to return a data frame with object names and types, but the stable version just expects a vector of table names.


javierluraschi:
LGTM.

PR #612, opened 2017-04-12
Not calling System.exit() inside the gateway thread in service mode

falaki:
sparklyr gateway thread calls System.exit when we come out of the infinite loop even when we are in service mode. This patch changes that. In both cases where I introduce a check against `isService`, not calling `System.exit` is harmless and the thread will exit.

### DIFF
--- a/java/backend.scala
+++ b/java/backend.scala
@@ -234,7 +234,7 @@ object Backend {
         System.exit(1)
     }
 
-    System.exit(0)
+    if (!isService) System.exit(0)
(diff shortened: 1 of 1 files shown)


kevinushey:
@javierluraschi, can you review when you're back in?

javierluraschi:
LGTM, merging.

In retrospect, we could jave removed `if (!isService) System.exit(0)` alltogether, but this makes the change scoped to service mode.

PR #614, opened 2017-04-12
fixes livy_service_start() when version is not specified

kevinykuo:
...but `spark_version` is.

```
# > livy_service_start(spark_version = "2.1.0")
# Error in if (version == "0.2.0" && numeric_version(spark_version) >= "2.0.0") { : 
#     missing value where TRUE/FALSE needed
```


javierluraschi:
Thank you!

PR #618, opened 2017-04-14
Add split separator pattern for R deser in Livy connections

kevinykuo:
This change adds `r_deser` to the return value of `split_separator()` for Livy so string columns deserialize back to R correctly

### DIFF
--- a/R/sdf_wrapper.R
+++ b/R/sdf_wrapper.R
@@ -49,7 +49,7 @@ sdf_deserialize_column <- function(column, sc) {
   separator <- split_separator(sc)
 
   if (is.character(column)) {
-    splat <- strsplit(column, separator$r, fixed = TRUE)[[1]]
+    splat <- strsplit(column, separator$r_deser, fixed = TRUE)[[1]]
     splat[splat == "<NA>"] <- NA
     Encoding(splat) <- "UTF-8"
     return(splat)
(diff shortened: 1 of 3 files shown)


R/sdf_wrapper.R
```diff
@@ -49,7 +49,7 @@ sdf_deserialize_column <- function(column, sc) {
   separator <- split_separator(sc)
 
   if (is.character(column)) {
-    splat <- strsplit(column, separator$r, fixed = TRUE)[[1]]
+    splat <- strsplit(column, separator$r_deser, fixed = TRUE)[[1]]
```
javierluraschi: Good catch!

However, the logic (not the behavior) is wrong for non-Livy connections as well, we should have use `separator$scala` here, this works fine for non-livy code, but not from livy. The separator named `scala` is used to deserialize the incoming data from scala using a regular expression. The separator named `r` is used to serialize data into scala (no need for regular expression). As you can see, `scala` and `r` and not very descriptive names in the first place.

So, what we would want to do here is rename `scala` into `regexp` and `r` into `plain`, or something of the likes and use the `regexp` version here and rename the instances of `scala` and `r` in `sdf_wrapper.R` and `data_copy.R` into `regexp` and `plain` respectively.

Mentioning this in case you have spare cycles to send another PR to improve the naming, but I think this is good to merge, thanks again for fixing this!

PR #620, opened 2017-04-14
Improve separator naming

kevinykuo:
Addresses comments in https://github.com/rstudio/sparklyr/pull/618

Also reverts a revert of a fix from my previous PR (noob git mistake!)


kevinushey:
Thanks!

PR #639, opened 2017-04-25
Feature/cbind

kevinykuo:
This change adds the `cbind()` method for `tbl_spark`.

Just starting out in Scala so the mechanism can probably be made more efficient. Feedback welcome!


javierluraschi:
@kevinykuo LGTM. Would you mind also adding an entry to https://github.com/rstudio/sparklyr/blob/master/NEWS.md?

kevinushey:
Would you be willing supply some unit tests as well?

kevinykuo:
Ready for review.

PR #641, opened 2017-04-26
Update README.Rmd

dgruenew:
Small (but significant) correction to `livy` connection setup


javierluraschi:
@dgruenew that section of the `README` does not need to trigger `spark_connect` twice. Could you give us more details as to why you are proposing this change `spark_connect`?

javierluraschi:
@dgruenew could we discuss this change in a github issue first? Thank you!

PR #643, opened 2017-04-27
ensure_scalar_* respects allow.null arg

### LINKED ISSUE #642: ensure_scalar_*() with NULL input and allow.null = TRUE throws error
Not clear if this is the desired behavior, but it's certainly unintuitive. 

To reproduce:

```r
ensure_scalar_character(NULL, allow.null = TRUE)
## Error in sparklyr::ensure_scalar_character(NULL, allow.null = TRUE) : 
##  'NULL' is not a length-one character vector
```

I see the same behavior with `ensure_scalar_boolean()` and the other functions created via `make_ensure_scalar_impl()`.

kevinykuo:
This change moves the logic for checking `NULL` up in the function returned by `make_ensure_scalar_impl()`. Fixes #642

`ensure_not_null()` seems redundant to the `checker(object)` condition so I removed it.

### DIFF
--- a/R/precondition.R
+++ b/R/precondition.R
@@ -61,6 +61,7 @@ make_ensure_scalar_impl <- function(checker,
 
     if (is.na(object)) object <- NA_integer_
     if (!allow.na)     ensure_not_na(object)
+    if (!allow.null)   ensure_not_null(object)
 
     converter(object)
   }
--- a/R/precondition.R
+++ b/R/precondition.R
@@ -54,12 +54,13 @@ make_ensure_scalar_impl <- function(checker,
   {
     object <- object %||% default
 
+    if (allow.null && is.null(object)) return(object)
+
     if (!checker(object))
       stopf("'%s' is not %s", deparse(substitute(object)), message)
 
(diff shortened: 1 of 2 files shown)


R/precondition.R
```diff
@@ -54,12 +54,13 @@ make_ensure_scalar_impl <- function(checker,
...
     object <- object %||% default
 
+    if (allow.null && is.null(object)) return(object)
+
     if (!checker(object))
       stopf("'%s' is not %s", deparse(substitute(object)), message)
 
     if (is.na(object)) object <- NA_integer_
     if (!allow.na)     ensure_not_na(object)
-    if (!allow.null)   ensure_not_null(object)
```
javierluraschi: Don't we need to keep this line? It's possible that all converters already throw an error on NULL, but since I'm not 100% sure, wouldn't hurt to keep this line, right?

kevinykuo: @javierluraschi Definitely doesn't hurt!

kevinushey:
Looks good to me -- thanks for the PR!

PR #658, opened 2017-05-05
Allow for splitting / extraction of elements from vector-columns (list-columns)

kevinushey:
This PR introduces a couple of helper functions for extracting elements from vector-columns -- typically, those returned by e.g. `sdf_predict()` or otherwise. See the test file for example usages.

I'm not entirely happy with the function names -- we should think about whether we could get something a bit more descriptive.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,9 +42,9 @@ Imports:
     shiny (>= 1.0.1),
     withr
 Remotes:
-    hadley/dplyr,
-    hadley/dbplyr,
-    hadley/rlang
+    tidyverse/dplyr,
+    tidyverse/dbplyr,
+    tidyverse/rlang
 Suggests:
     testthat,
     RCurl,


javierluraschi:
LGTM. I don't have better suggestions for naming `sdf_extract_column` and `sdf_split_column`, both seem reasonable to me.

kevinykuo:
We'll have both `sdf_separate_column()` and `separate.tbl_spark` (from tidyr)?

kevinushey:
Right -- for now, we'll just have a separate `sdf_separate_column()` analog but we'll also provide a method for `tidyr` if possible. The interface for `tidyr::separate()` is a bit more complicated so I wanted to get a simple standalone version in first if possible.

kevinushey:
Note to self: before merging, we should make sure this works with e.g. the return value of tokenizers.

Issue #644, opened 2017-04-27
sdf_bind_rows and sdf_bind_cols


kevinykuo:
Any disadvantages to implementing/exporting these now before `bind_rows()` and `bind_cols()` are made generic in dplyr?

kevinykuo:
@javierluraschi let me know whenever you'd welcome a PR. This would address https://github.com/rstudio/sparklyr/issues/556 and https://github.com/rstudio/sparklyr/issues/507 for the time being.

PR #662, opened 2017-05-05
Implement sdf_bind_rows and sdf_bind_cols

kevinykuo:
**Ready for review**

This PR would add `sdf_bind_rows` and `sdf_bind_cols` which mirror `dplyr::bind_rows` and `dplyr::bind_cols`.

I need some help with the `sdf_bind_cols` implementation: the issue relates to our calling `substitute()` in `cbind.tbl_spark()` in order to generate an informative message when there's a mismatch in number of rows (https://github.com/rstudio/sparklyr/blob/dad79603b3b9779930283411061c85be7e93aeac/R/mutation.R#L56-L68)

The behavior in this commit:

```
df1a <- data_frame(a = 1:3, b = letters[1:3])
df3a <- data_frame(e = 1:2, f = letters[1:2])

df1a_tbl <- copy_to(sc, df1a)
df3a_tbl <- copy_to(sc, df3a)

cbind(df1a_tbl, df3a_tbl)
# Error in cbind(deparse.level, ...) : 
#   Not all inputs have the same number of rows, for example:
#   tbl nrow
# 1 df3a_tbl    2
# 2 df1a_tbl    3 

sdf_bind_cols(df1a_tbl, df3a_tbl)
# Error in cbind(deparse.level, ...) : 
#   Not all inputs have the same number of rows, for example:
#   tbl nrow
# 1   2    2
# 2   1    3 
```

To make this work, we'd need to ship the expressions passed to `sdf_bind_cols()` through the call to `cbind()`. This is made more complicated since we need to support lists of `tbl_spark`. There should be some `rlang` tricks available to make this work, but I can't wrap my head around it.

Any comments appreciated with this or other parts of the PR.


kevinykuo:
Played around some more and this is what I could come up with. R is hard...

```
sdf_bind_cols <- function(...) {
  dots <- Filter(length, rlang::dots_splice(...))
  if (! all(sapply(dots, is.tbl_spark)))
    stop("all inputs must be tbl_spark")

  nonempty <- rlang::dots_splice(...) %>%
    sapply(Negate(rlang::is_empty)) %>%
    which()
  filtered_exprs <- rlang::dots_exprs(...) %>%
    lapply(function(x) if (rlang::is_lang(x)) rlang::lang_tail(x) else x) %>%
    unlist() %>%
    `[`(nonempty)

  rlang::lang("cbind", !!! filtered_exprs)  %>%
    rlang::eval_tidy()
}

# this works
sdf_bind_cols(df1a_tbl, list(df3a_tbl))
# Error in cbind(deparse.level, ...) : 
#   Not all inputs have the same number of rows, for example:
#   tbl nrow
# 1 df3a_tbl    2
# 2 df1a_tbl    3 

# but then of course this wouldn't
myfun <- function(x, y) {
  sdf_bind_cols(x, y)
}
myfun(df1a_tbl, df3a_tbl)
# Error in cbind(x, y) : object 'x' not found
```

kevinushey:
`cbind()` and `rbind()` are notoriously thorny when it comes to implementing custom methods. Because dispatch occurs only on the first argument, but users could conceivably attempt to bind differing types of objects together.

I'll try to take a deeper look this week.

kevinykuo:
I think I got it, tidyeval is much fun. Ready for review when y'all get a chance.

javierluraschi:
@kevinykuo this is great!

I was thinking that, ideally, we want to override `dplyr::bind_cols` and `dplyr::bind_rows` since it looks that you are motivated to bring parity to those functions. However, I don't think there is a clean way to override this yet, so I've opened this issue https://github.com/tidyverse/dplyr/issues/2759. Once `dplyr` provides them as `S3` methods, we can implement them using `sdf_bind_rows` and `sdf_bind_cols`.

kevinykuo:
@javierluraschi Agreed! See https://github.com/tidyverse/dplyr/issues/2457. I wasn't sure when the `dplyr` folks will get to it so I figured we'd start now (https://github.com/rstudio/sparklyr/issues/644)

javierluraschi:
Thanks for chatting with Hadley about this, once it becomes available in `dplyr`, we can implement them using `sdf_bind_rows` and `sdf_bind_cols`.

kevinykuo:
Ugh. I need git therapy.

Issue #661, opened 2017-05-05
Consider rebasing sdf_with_sequential_id from 1 instead of 0


kevinykuo:
See https://github.com/rstudio/sparklyr/issues/652

kevinushey:
Making the default base be 1 sounds like the right idea to me, but we might also want to add `base` as an argument to the function, just to allow users to control where indexing starts (in case that's important for whatever reason)

PR #664, opened 2017-05-05
add 'from' argument to sdf_with_sequential_id and defaults to 1

kevinykuo:
This change adds a `from` argument to `sdf_with_sequential_id()` and defaults it to `1L`.

Closes https://github.com/rstudio/sparklyr/issues/661.


kevinushey:
LGTM!

PR #672, opened 2017-05-10
Implement residuals() for Spark ML linear and generalized linear regressions

kevinykuo:
This PR implements `residuals()` for `ml_model_generalized_linear_regression` and `ml_model_linear_regression`. It removes `residuals.ml_model_logistic_regression` and instead uses  `residuals.ml_model` to inform users that the method isn't implemented yet.

These methods have a `collect` argument which determines whether to convert the `DataFrame` to a R vector. It should be set to `FALSE` for very large datasets (which returns a `tbl_spark`) but defaults to `TRUE` for compatibility.

For `residuals.ml_model_generalized_linear_regression()`, the default type is `deviance` to be consistent with `residuals.glm`.


R/ml_generalized_linear_regression.R
```diff
@@ -210,3 +210,23 @@ summary.ml_model_generalized_linear_regression <-
...
+
+  type <- rlang::arg_match(type)
+  ensure_scalar_character(type)
+  stopifnot(is.logical(collect))
+
+  residuals <- object$.model %>%
+    invoke("summary") %>%
+    invoke("residuals", type)
+
+  if (collect == TRUE)
```
kevinushey: Minor, but the `== TRUE` here is unnecessary (can just write `if (collect)`

R/ml_generalized_linear_regression.R
```diff
@@ -210,3 +210,23 @@ summary.ml_model_generalized_linear_regression <-
...
+
+#' @export
+residuals.ml_model_generalized_linear_regression <- function(
+  object,
+  type = c("deviance", "pearson", "working", "response"),
+  collect = TRUE, ...) {
+
+  type <- rlang::arg_match(type)
+  ensure_scalar_character(type)
+  stopifnot(is.logical(collect))
```
kevinushey: Why not `ensure_scalar_boolean()` here?

kevinushey:
I kind of wonder if we should instead adopt the convention that `residuals()` always returns an R vector (ie it always collects the result), and `sdf_residuals()` or similar returns a Spark DataFrame with the residuals added as a column.

kevinushey:
Also -- I worry that the attempt to compute residuals might mutate the row-order of the Spark Dataset, thereby making the `sdf_read_column()` return values incorrect (they could be in the wrong order). We might need to do something similar as per:

https://github.com/rstudio/sparklyr/blob/75ad74bb50a3829f64cbfc8c8dc7add91b464df7/R/ml_utils.R#L97-L126

In other words, I think our implementation for `residuals()` should mirror what we have for `predict()`.

kevinykuo:
Closing for now. Will look into implementing a general way to add residuals columns as discussed.

PR #673, opened 2017-05-10
add ml_model_data()

### LINKED ISSUE #555: Add a method to get the data from a model
It's sometimes useful to be able to extract the data from model objects returned by `ml_*` functions. It currently takes a bit of digging to figure out what to do.

Here's my suggestion:

```r
tbl.ml_model <- function(x) {
  sdf_register(x$data)
}
```

Then you can just get the data in the natural way, using `tbl(a_model)`.

kevinykuo:
Adds method to extract training data for a Spark ML model. Closes #555


javierluraschi:
LGTM. Update to `NEWS`?

Issue #670, opened 2017-05-09
Implement broom functions


kevinykuo:
Working on this. Will start with GLM type models. Blocked by https://github.com/rstudio/sparklyr/issues/557 but the GLM case is easy to resolve so I'll submit a fix for that soon.

TODO
- [x] Implement `residuals()` for GLM type models
- [x] Implement method to get training data from model https://github.com/rstudio/sparklyr/issues/555
- [x] Implement `tidy`, `augment`, and `glance`
- [x] Informative error messages for models not yet supported
- [x] Tests
- [x] Support logistic regression and rest of supervised learning models
- [x] Support unsupervised learning models

kevinykuo:
Regarding column names in `broom` output -- should we follow the naming convention with dots, e.g. `std.error` and `.resid`, or come up with our own scheme since we've sort of gone down the path of sanitizing column names?

@kevinushey

kevinushey:
That's a good question... from the README at https://github.com/tidyverse/broom, it sounds like the convention is to use `.`-separated names, so we should likely want to use that here. However, given that some versions of Spark don't handle `.`s in names all that well, we should provide a mechanism for users to generate names using camelCase as well.

What's your opinion here?

kevinushey:
Maybe we want something like `tidy(data, style = "dotcase")` vs. `tidy(data, style = "camelcase")`?

kevinykuo:
I'm actually leaning towards `dot.case` for broom consistency and `snake_case`, because tidyverse. Will continue this discussion in a PR.

yitao-li:
Looks like this work is actually complete by now. We can close this issue.

Issue #557, opened 2017-03-15
Add residuals() method for ml_model objects


richierocks:
At the moment, it's slightly tricky to get the residuals from a model.

The basic idea is this:

```r
residuals.ml_model <- function(x, ...) {
  sdf_register(x$data) %>% 
    select_(actual = x$response) %>% 
    transmute(
      residual = predict(x) - actual
    )
}
```

This doesn't work for me without `collect()`ing the data, and for some models you need to think about weights, but the basic idea is right.

kevinushey:
We do in fact have a `residuals.ml_model()` method already; it's defined like so:

```
#' @export
residuals.ml_model <- function(object, ...) {
  object$.model %>%
    invoke("summary") %>%
    invoke("residuals") %>%
    sdf_read_column("residuals")
}
```

This may perhaps only be valid for e.g. Spark linear regression models, though. Can you elaborate on what kind of model you're trying to get residuals from?

richierocks:
@kevinushey Oh, so you do. I tried `residuals()` on a random forest model, and got this error.

```
 Error: java.lang.IllegalArgumentException: invalid method summary for object 750
```

A method for gradient boosted trees would also be really useful.

kevinushey:
Gotcha, that makes sense. I agree it would be worth adding a more general `residuals()` implementation here.

PR #685, opened 2017-05-16
residuals() and sdf_residuals() for lm/glm

kevinykuo:
This PR implements `residuals()` and `sdf_residual()` methods for Spark ML linear regression and GLM regression models. The former returns a R vector and the latter returns a `tbl_spark`.

After looking at the MLlib code I discovered that it wasn't possible to simply "select" a residuals column *in addition* to the modeling DataFrame. The methods to calculate the ingredients for residuals (e.g. deviance) were not exposed in the API so we can't invoke them directly (or via reflection.) Instead of implementing a family/distribution hierarchy in sparklyr, I'm proposing that we "cbind" the output that's provided by the Spark API (i.e. a DataFrame with one column) with the the modeling dataset. Since the residuals output from Spark and the training dataset have the same distribution/order in the cluster (it's just a udf map), we can invoke `rdd.zip` which doesn't involve shuffling. This mechanism is implemented in the unexported function `sdf_fast_bind_cols()`.

The naming of the output column `"residuals"` was chosen to be consistent with `"predictions"` in `sdf_predict()`.

Related: https://github.com/rstudio/sparklyr/issues/670, https://github.com/rstudio/sparklyr/issues/557

Previous PR: https://github.com/rstudio/sparklyr/pull/672


R/sdf_utils.R
```diff
@@ -0,0 +1,32 @@
...
+#' This is a version of `sdf_bind_cols` that works by zipping
+#' RDDs. From the API docs: "Assumes that the two RDDs have the
+#' *same number of partitions* and the *same number of elements
+#' in each partition* (e.g. one was made through a map on the
+#' other)."
+#'
+#' @keywords internal
+#'
+#' @param ... Spark DataFrames to cbind
+sdf_fast_bind_cols <- function(...) {
```
kevinushey: Is it possible for us to validate the requirement above re: two RDDs having same number of partitions + same number of elements in each partition? (Just to future-proof us in case we accidentally use this function where the above doesn't hold)

R/sdf_utils.R
```diff
@@ -0,0 +1,32 @@
...
+    spark_dataframe() %>%
+    spark_connection()
+
+  zip_sdf <- function(df1, df2) invoke_static(sc,
+                           "sparklyr.Utils",
+                           "zipDataFrames",
+                           spark_context(sc),
+                           df1, df2)
+
+  Reduce(zip_sdf, lapply(dots, spark_dataframe)) %>%
```
kevinushey: Nice use of `Reduce` here :)

R/sdf_utils.R
kevinykuo: Checking the number of partitions is easy. Counting elements in each partition is more expensive since we have to iterate through them (albeit in parallel.) Maybe we can have a `validate` argument in the function that defaults to `TRUE`?

kevinushey: I think that sounds good. Alternatively, we could hide the validation code behind some R / sparklyr configuration option?

Either way, maybe let's leave that for a separate PR in case we find that we want to use `sdf_fast_bind_cols()` in other contexts where it's less obvious that the required preconditions are met (for some reason).

PR #701, opened 2017-05-25
Implement broom for ML glm and linear regression models

kevinykuo:
This PR adds implementations of `tidy`, `augment`, and `glance` for ML GLM and linear regression models (no logistic regression yet.)

Design decisions:

- @kevinushey We discussed whether to include a `style` argument so users could choose between `dot.case` or something else for column names since some versions of Spark don't support `.` in column names. I realized this wasn't necessary because the only methods this would affect are `tidy` and `glance`, whose outputs are always small `data.frame`. `broom` chose to prefix names of added columns in `augment` with dots, but we can't have column names starting with dots in Spark anyway. In my `augment` output, the added columns are simply named `fitted` and `resid`.

- The `type.residuals` argument to `augment` defaults to `"working"`, i.e. `actual - predicted` or prediction error. For GLM, this is different from the default for `residuals()` and `sdf_residuals()`, which is `"deviance"` (to be consistent with base R methods.) This default is chosen because 1) we don't support anything other than "working" when `newdata` is supplied, so it would avoid confusion, also 2) IMO most people think prediction error when they see "residuals" in the ML context.

- Working residual is defined to be `actual - predicted` instead of `predicted - actual`; just agreeing with base R and MLlib here.

Related: https://github.com/rstudio/sparklyr/issues/670, https://github.com/rstudio/sparklyr/issues/557


kevinykuo:
Build failure unrelated to PR:

```
checking whether package ‘sparklyr’ can be installed ... WARNING
Found the following significant warnings:
  Warning: Installed Rcpp (0.12.11) different from Rcpp used to build dplyr (0.12.10).
```

javierluraschi:
Let me rebuild form travis, just in case.

kevinykuo:
Lemme try rebase/push.

kevinykuo:
Something funky with the broom import?

javierluraschi:
Seems the passing one is using `glue 1.0.0.9000` and `curl 2.6`:

screen shot 2017-05-25 at 9 50 38 pm

The failing one is using `glue 1.0.0` and `curl 2.5`:

screen shot 2017-05-25 at 9 50 29 pm

It doesn't seem this is an issue with the PR since both share the same commit... probably a caching issue in travis making `curl` stuck on `2.5`. I'll merge but will have to revert and help investigate further if this breaks.

javierluraschi:
@jimhester does this issue sound familiar? Looks to me like, somehow, Travis is using an older version of `curl` in some workers since the same commit was run in two different repos and the environment seems different between them, see side-by-side comparison above. Thank you!

javierluraschi:
@kevinykuo could you reopen this PR? I just rebased this https://github.com/rstudio/sparklyr/pull/698 and didn't hit this error, maybe rebasing over these changes will fix this issue? Otherwise, I bet Jim can give us a couple hints tomorrow.

PR #702, opened 2017-05-25
Implement checkpoint directory and checkpointing

kevinykuo:
Setting a checkpoint directory is required for some algorithms e.g. LDA and GraphFrames Connected Components.


javierluraschi:
Rebuild jars? Otherwise LGTM.

kevinykuo:
@javierluraschi OK should be good now.

PR #708, opened 2017-05-26
Add broom to DESCRIPTION

javierluraschi:
@kevinykuo I wonder if the order of `broom` is causing this issue. BTW. The problem seems to be that `broom` depends on `dplyr`/`dbplyr` from CRAN not the development versions we are using in `sparklyr`. Lets see if this works...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,5 @@
 language: r
 sudo: false
 cache: packages
+r_github_packages:
+ - hadley/devtools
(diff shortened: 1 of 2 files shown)


kevinykuo:
I was actually trying exactly this but didn't want to push it since the non-alphabetic order drove me a little crazy :P

javierluraschi:
This didn't work but at least we know that adding `broom` is what is causing this, trying now with `devel` version to match `dplyr`...

javierluraschi:
@kevinykuo there you go! I guess using `tidyverse/broom` avoids importing the `dplyr` CRAN version. We should be able to merge after this change, I'll close this PR.

PR #709, opened 2017-05-26
broom for ml glm/lm

kevinykuo:
Hope it works this time!


javierluraschi:
Shoot, on my PR I only added `widyverse/broom` as a remote not in the imports... you have the right PR but now we don't know if this will actually fix it. :crossed_fingers:

Issue #688, opened 2017-05-17
Implementing functionality not available in older Spark versions


kevinykuo:
If we write a function in e.g. `utils.scala` that depends on a Spark function not available before `2.0.0`, it won't compile when we build against `1.6`. We probably don't want to *not* implement newer features just because we support older versions. What's the best way to handle this and what should we do in the meantime?

kevinushey:
Although we don't use this yet, my initial thought was to enable this through the use of a filter in the compilation spec:

https://github.com/rstudio/sparklyr/blob/d7ccc2e10a84eb69276a66c1bdf623ba645859c0/R/spark_compile.R#L50-L52

https://github.com/rstudio/sparklyr/blob/d7ccc2e10a84eb69276a66c1bdf623ba645859c0/R/spark_compile.R#L229-L263

The idea there being, we could add some custom filters that add / remove Scala files according to the version of Spark being compiled against.

I'm still not sure what the best way of organizing this would be though; e.g. do we want to have version-specific folders that add custom code for a particular version of Spark? E.g. what's some convention we could adopt that says, "use this code when compiling against Spark 2.0 only, use this code when compiling against Spark 2.0 and above", and so on.

kevinykuo:
Since Scala doesn't support partial classes (i.e. splitting a class into different files), I wonder if we can do something like the following:

Have a `java/spark-1.6` folder with the current Scala code, but change objects to traits and append the version associated, e.g. `trait Utils16`. Then we'd have the same structure in `java/spark-2.0` but it would only contain the "diff" to include new methods. Now we'd have files under `java` for object declarations, e.g. `spark16.scala` might say `object Utils extends Utils16` and `spark20.scala` might say `object Utils extends Utils16 with Utils20`.

In the compilation spec we'd compile the folders with spark version up to the one each jar is built against, plus the file that declares the objects.

This assumes that we'll be able to resolve any breaking changes to the Spark API from one version to the next.

kevinykuo:
@javierluraschi @kevinushey thoughts?

kevinushey:
We could also just perform the indirection at the R level -- have a bunch of different versioned `Utils` objects, and figure out which one should be used at runtime with R, rather than dealing with Scala inheritance. Does that make sense?

In other words, from R, we just explicitly call e.g.

```
invoke_static(sc, "sparklyr.UtilsV2", ...)
```

[...]

Issue #714, opened 2017-05-30
Implement sdf_repartition


kevinykuo:
I think I have the PR for this, but am blocked by https://github.com/rstudio/sparklyr/issues/688

kevinykuo:
Popped up in my SO feed: https://stackoverflow.com/questions/44251185/how-to-repartition-a-data-frame-in-sparklyr

PR #719, opened 2017-06-01
Support implementation of Spark 2.0+ features and add sdf_repartition()

kevinykuo:
This change

- Refactors the directory structure of Scala code to support implementation of newer Spark functionality.
- Utilizes the `scala_filter` in default compilation configuration to filter appropriate source code for various Spark versions.
- Implements `sdf_repartition()` which relates to changes above (due to different signatures in Spark 1.6 vs 2), hence the inclusion in this PR.

Proposed directory structure:
- java
    - spark-1.5.2
        - utils.scala
        - (all other current .scala files)
    - spark-2.0.0
        - utils200.scala

Closes: https://github.com/rstudio/sparklyr/issues/688, https://github.com/rstudio/sparklyr/issues/714


R/sdf_interface.R
```diff
@@ -694,3 +694,36 @@ sdf_broadcast <- function(x) {
...
+#'
+#' @param partitions number of partitions
+#' @param columns vector of column names used for partitioning, only supported for Spark 2.0+
+#'
+#' @export
+sdf_repartition <- function(x, partitions = NULL, columns = NULL) {
+  sdf <- spark_dataframe(x)
+  sc <- spark_connection(sdf)
+
+  partitions <- partitions %||% 0L
```
kevinushey: Maybe I'm misunderstanding something, but it looks like we don't allow `NULL` here (we remap that to `0L`) -- so is the `allow.null` below needed?

R/spark_compile.R
```diff
@@ -45,11 +45,13 @@ spark_compile <- function(jar_name,
...
   scala_path <- file.path(root, "java")
-  scala_files <- list.files(scala_path, pattern = "scala$", full.names = TRUE)
+  scala_files <- list.files(scala_path, pattern = "scala$",
+                            full.names = TRUE,
+                            recursive = TRUE)
 
   # apply user filter to scala files
   if (is.function(filter))
-    scala_files <- filter(scala_files)
+    scala_files <- Filter(filter, scala_files)
```
kevinushey: Is there a reason why you want to use `Filter` here? My thought is that the filter function here would take a vector of scala files, and return a new vector of scala files. It seems like we don't really gain much by hiding the vectorization.

R/spark_compile.R
```diff
@@ -367,3 +373,17 @@ get_scalac_version <- function(scalac = Sys.which("scalac")) {
   splat <- strsplit(version_string, "\\s+", perl = TRUE)[[1]]
   splat[[4]]
 }
+
+make_version_filter <- function(version_upper) {
```
kevinushey: Since this is a function that returns a function, to be safe, you'll need to call `force(version_upper)` before the returned function definition.

kevinushey:
The moving around of the older Scala source files might affect the scripts we use to copy these for `livy` -- @javierluraschi, can you take a look and double check this new scheme is compatible there?

java/spark-2.0.0/utils200.scala
```diff
@@ -0,0 +1,19 @@
+package sparklyr
```
kevinushey: Minor, but can we just call this file `utils.scala`? (It's already implicitly versioned as it's part of the 2.0.0 folder). Or does the file name need to be the same as the object name for Scala?

kevinykuo: We can call it `utils.scala` but it'd mean we can have a couple `utils.scala` tabs in RStudio. I actually prefer `utils.scala` for the reason you mentioned, and can change it back if the duplicate name in the code editor doesn't bother us.

R/sdf_interface.R
kevinykuo: Yep the `allow.null` is superfluous. Will fix.

kevinykuo:
Yeah looks like livy is breaking :(

```
> iris_tbl %>%
+   sdf_repartition(columns = c("Species", "Petal_Width"))
 Show Traceback
 
 Rerun with Debug
 Error: java.lang.ClassNotFoundException: Utils200
	at scala.reflect.internal.util.AbstractFileClassLoader.findClass(AbstractFileClassLoader.scala:62)
```

R/spark_compile.R
kevinykuo: I figured it's slightly easier to provide a function that handles one file, but given that this has been an exported function I probably shouldn't have messed with it. Will revert and modify `make_version_filter()`.

java/spark-2.0.0/utils.scala
```diff
@@ -0,0 +1,19 @@
+package sparklyr
+
+import org.apache.spark.sql._
+
+object Utils200 {
```
javierluraschi: @kevinykuo could we rename this object to `Repartition`, `RepartitionUtils` or something like that? I think we should avoid versioned if possible.

From the code, it looks to me like we are compiling `Utils.scala` for Spark 1.X, could you consider using `scala_filter` under [spark_default_compilation_spec](https://github.com/rstudio/sparklyr/blob/8da6b058848ae9a3a47b95bd981bc509032f02c6/R/spark_compile.R#L236-L263) to avoid compiling any `2.x` files under `1.x`?

javierluraschi:
@kevinykuo to fix Livy, you probably need to fix the paths under [livy_sources_included](https://github.com/rstudio/sparklyr/blob/ce024629e84d918faf94c2de2d61bfa6f86e3f48/R/livy_sources.R#L1-L10).

FYI. We don't do much for "compiling" Livy sources, mostly, we strip out a few headers and make them available inside the package to push the through Livy when a connection starts.

R/spark_compile.R
```diff
@@ -239,25 +241,29 @@ spark_default_compilation_spec <- function(pkg = infer_active_package_name()) {
...
+      jar_path = find_jar(),
+      scala_filter = make_version_filter("1.5.2")
     ),
     spark_compilation_spec(
       spark_version = "1.6.1",
       scalac_path = find_scalac("2.10"),
       jar_name = sprintf("%s-1.6-2.10.jar", pkg),
-      jar_path = find_jar()
+      jar_path = find_jar(),
+      scala_filter = make_version_filter("1.6.1")
```
kevinykuo: @javierluraschi `scala_filter` is indeed used, otherwise the Scala compiler would complain.

java/spark-2.0.0/utils.scala
kevinykuo: @javierluraschi the thought here was that there may be more new functions down the road in later versions of Spark, and we might not want to create a new object for each function, hence the versioning. However, it seems like the collection of methods have been stable enough, so we're probably OK calling this `Repartition`. You're fine with having versioned directory names though?

javierluraschi: Yes, I think versioned directories is the right approach, we've been lucky so far to be able to workaround versioning issues in a single set of sources, but we were expecting this to happen at some point.

R/spark_compile.R
javierluraschi: Ah, great!

javierluraschi:
Nice one @kevinykuo, merged.

PR #724, opened 2017-06-04
add DBIConnection to livy sc and implement create_hive_context for livy

kevinykuo:
@javierluraschi here's an attempt to get the `sdf_` functions to work


javierluraschi:
LGTM

PR #725, opened 2017-06-04
add sdf_num_partitions

kevinykuo:
This PR adds `sdf_num_partitions()` for getting the number of partitions of a Spark DataFrame. I think it's useful and both PySpark and SparkR have it.


javierluraschi:
Looks good!

PR #728, opened 2017-06-05
Implement R Workers

javierluraschi:
This PR implements support for executing distributed R operations under a new R worker role and adds support for `spark_apply` to transform data frames row-by-row over an arbitrary closure.

## Overview

screen shot 2017-06-05 at 9 36 40 am

The current architecture of `sparklyr` relies on two components:
1. The **sparklyr client**: Composed of source files under [/r/](https://github.com/rstudio/sparklyr/tree/master/R) which provide an interface to Spark by establishing and using a connecting to the backend.
2. The **sparklyr backend**: Composed of sources under [/scala/](https://github.com/rstudio/sparklyr/tree/master/java) which provide support to proxy commands from the client into Spark.

So far, this has been enough to provide analysis, machine learning and extensibility capabilities to `sparklyr` since there is a rich set of functionality already implemented into Spark and provided by the Spark community. Additionally, one could write extensions in `sparklyr` with Scala code that would enable arbitrary functionality.

However, there are still cases where users would prefer or might require executing specific functionality from R at scale in Spark. This PR implements enable this functionality by providing:
1. Support for `WorkerRDD` that extends Apache Spark `RDD`'s. 
2. Support for `spark_apply` to compute a function closure that serializes to the workers.
2. Support for starting an `Rscript` in each `RDD` partition.
3. Support for starting an `sparklyr` backend in each `RDD` partition.
4. Support for connecting back to the backend one the `rscipt` initializes.
5. Support to retrieve the partition, apply the closure and update the results.

Notice that a new class of files was introduced prefixed:
1. **worker_**: This prefix indicates that this is code meant to be executed in the worker node with the main function defined as `spark_worker_main`.
2. **core_**: This prefix indicates shared functionality between client and worker. It was explicitly separated under `core_` since absolutely no package dependency can be assumed in this files. So for instance, use of the `config`, `yaml`, `assertthat` or any other package would result in the worker nodes from failing.

## Examples

First, some standard initialization:

```{r}
library(sparklyr)
library(dplyr)

sc <- spark_connect(master = "local", version = "2.1.0")
iris_tbl <- copy_to(sc, iris)
```

A common distributed computation example is to approximate the value of π. In this case, we will compute whether a point is inside the circle using an R function, but then use `dplyr` to filter and aggregate the results of this approximation:

```{r paged.print=F}
sdf_len(sc, 10000) %>%
  spark_apply(function() sum(runif(2, min = -1, max = 1) ^ 2) < 1) %>%
  filter(id) %>% count() %>% collect() * 4 / 10000
```

```
  n
1 3.13
```

Another example is to apply an R specific function,  say`rgamma`, over the a dataset as follows:

```{r}
spark_apply(iris_tbl, function(row) {
  lapply(row[1:4], function(e) e + rgamma(1,2))
})
```

```
# Source:   table<sparklyr_tmp_d34536c5db4c> [?? x 4]
# Database: spark_connection
   Sepal_Length Sepal_Width Petal_Length Petal_Width
          <dbl>       <dbl>        <dbl>       <dbl>
 1     7.676218    6.407721     3.993295   1.4467454
 2    10.373125    5.819972     2.933998   1.7986557
 3     5.648909    6.097531     3.322528   2.6482938
 4     6.807209    3.898670     3.775738   1.9602965
 5     9.943737    5.164587     2.282128   1.1904444
 6     9.240930    4.492142     3.866824   1.7283005
 7     7.814709    3.776937     1.987202   3.9390395
 8     7.257176    5.386991     3.768895   3.3914439
 9     5.921331   13.847358     2.880484   0.2565747
10     5.993263    4.615195     3.334328   0.8899092
# ... with 140 more rows
```

Notice that `spark_apply` takes the column names of the output data frame but infers them from `colnames(x)`. This works well in the examples above, but one might want to explicitly name the output as follows:

```{r}
spark_apply(iris_tbl, function(row) {
  row$Petal_Width + rgamma(1,2)
}, names = "randomized_petal_width")
```

```
# Source:   table<sparklyr_tmp_d34577c076ab> [?? x 1]
# Database: spark_connection
   randomized_petal_width
                    <dbl>
 1               2.904409
 2               1.788410
 3               1.576618
 4               2.588035
 5               2.808169
 6               1.611326
 7               1.804285
 8               1.106392
 9               3.971133
10               2.848165
# ... with 140 more rows
```

One can also retrieve multiple columns in each row by returning a list, and explicitly name the columns in output data frame:

```{r}
spark_apply(iris_tbl, function(row) {
  list(row$Petal_Width + rgamma(1,2), 1)
}, names = list("random", "ones"))
```

```
# Source:   table<sparklyr_tmp_d3457caf11f8> [?? x 2]
# Database: spark_connection
      random  ones
       <dbl> <dbl>
 1 1.7492575     1
 2 0.7582578     1
 3 1.3671871     1
 4 2.6247821     1
 5 4.2132402     1
 6 0.6252375     1
 7 1.1091248     1
 8 1.7231689     1
 9 1.7179472     1
10 1.0698262     1
# ... with 140 more rows
```

## Next steps
1. **Support for packages**. Currently, closures need to be dependency free or alternatively, each worker node needs to be correctly configured with preinstalled libraries. It is obviously a priority to enable this scenario so my plan is to support this soon.
2. **Additional functions**. Other ways of computing arbitrary code that are not strict mappings say by filtering, grouping, etc. Future work will implement those.
3. **Efficient serialization**. It is obvious now that serialization across `sparklyr` client/backend could be significantly improved by using technologies like Apache Arrow or Feather.
4. **R daemon process**. Currently, the`rscript` is launched each time a partition requires computation which has a fixed initialization cost, this could be improved with longed-lived workers.


run-tests
```diff
@@ -1,16 +0,0 @@
-#!/usr/bin/env bash
```
kevinushey: Is there a reason why you're removing this? (I've been running this after commits to ensure they don't regress different versions of Spark installed)

R/core_gateway.R
```diff
@@ -0,0 +1,124 @@
+wait_connect_gateway <- function(gatewayAddress, gatewayPort, config, isStarting) {
+  waitSeconds <- if (isStarting)
+    spark_config_value(config, "sparklyr.gateway.start.timeout", 60)
+  else
+    spark_config_value(config, "sparklyr.gateway.connect.timeout", 1)
+
+  gateway <- NULL
+  commandStart <- Sys.time()
+
+  while (is.null(gateway) && Sys.time() < commandStart + waitSeconds) {
```
kevinushey: Does this loop need a `Sys.sleep()` call? (do we want to wait for some interval after a failed attempt to create a `socketConnection()`, or is it okay to eagerly keep trying?)

configure.R
```diff
@@ -1,4 +1,11 @@
 #!/usr/bin/env Rscript
 library(sparklyr)
 sparklyr:::livy_sources_refresh()
+
+sparklyr:::spark_compile_embedded_sources()
+
+javaopts <- Sys.getenv("CLASSPATH")
```
kevinushey: Should this be `JAVA_OPTS`?

kevinushey:
Is there some way of performing 'chunked' operations here? That is, rather than performing operations row-by-row, perform them in chunks (e.g. perhaps by partition) or similar?

It would be nice if there were some way to define the partitioning / chunking strategy.

kevinushey:
There's _got_ to be a better way of handling package dependencies / `sparklyr` sources than just generating that `sources.scala` monstrosity. Why can't the workers just use `sparklyr`, downloading and installing it on the fly if necessary?

run-tests
javierluraschi: No, thanks for pointing this out! I had a bunch of files that were no longer used in `sparkworker` and this one got mixed with them, reverted that commit.

configure.R
javierluraschi: Correct, changed. Thanks!

javierluraschi:
@kevinushey regarding the two remaining comments:

1. Sort of, let me take a look. In general, the RDD gives you an iterator over a ROW object; however, we are already serializing this as a chunk but not being exposed in `spark_apply`, if I can't get this done today, I'll open an issue to continue the investigation after the merge.

2. I very much agree. We can't do the package download since a good amount of clusters block internet connectivity. However, we should be able to pass a reference to `sparklyr` over `spark-submit` or as file that gets distributed to all workers. If we pass this after `spark-submit`, we would be basically also resolving this https://github.com/rstudio/sparklyr/issues/729 which tracks the ability to install packages during a call to `spark_apply`. I'll see what I can improve today but might have to follow up after the merge.

kevinushey:
Sounds good. I don't think any of my comments block merge of this PR (nice work!) but I just want to make sure we keep this in mind.

IMHO the primary interface we should aim for is one where we have R functions that accept an R data.frame, and return an R data.frame, and behind the scenes `sparklyr` manages the splitting + combining of the various data.frames received. (Ie, I'm imagining this as a 'split-apply-combine' model, where we take care of 'split' and 'combine' behind the scenes, and allow the user to supply some relatively arbitrary 'apply' function)

kevinykuo:
I agree that we need to support the "split-apply-combine" model. Many data science use cases would involve a pattern like this: start with a Spark DataFrame (say from a hive table), partition it based on some key, running complex transformations or models (via R) off each (reasonably sized) partition, then putting the results back in a Spark DataFrame or just collecting. Basically the `group_by-do`/`nest-mutate`/`SparkR::g{d}apply` type of patterns.

javierluraschi:
@kevinushey changed `spark_apply` from providing/expecting a data frames. I agree with you this is the right approach. Originally, I was thinking in terms of Spark Streaming which might return single-row data frames and the fact that we have a row iterator in Scala which maps better to single-rows, but everything in R is vectorized so makes sense to take the most out of this.

A few random comments...

Here are some performance numbers (for the rest of the community: we do want to improve significantly over them soon), for row-by-row:

```{r}
system.time({
spark_apply(iris_tbl, function(row) {
  lapply(row[1:4], function(e) e + rgamma(1,2))
}) %>% collect()
})

  0.161   0.039  11.205 
  0.056   0.037  11.108 
```

Changing this into a data frame (latest commit) does not change performance significantly:

```{r}
system.time({
  spark_apply(iris_tbl, function(data) {
    data[1:4] + rgamma(1,2)
  }) %>% collect()
})

user  system elapsed 
  0.053   0.035  11.128 
  0.057   0.036  11.119 
```

That said, most of the time is spend on serialization which we (read I) should improve soon. The change to data frames does simplify the closure significantly from `lapply(row[1:4], function(e) e + rgamma(1,2))` ti `data[1:4] + rgamma(1,2)`, looks nice!

However, for the Pi approximation closure:

```{r}
system.time({
sdf_len(sc, 10000) %>%
  spark_apply(function() sum(runif(2, min = -1, max = 1) ^ 2) < 1) %>%
  filter(id) %>% count() %>% collect() * 4 / 10000
})

user  system elapsed 
  0.107   0.053  17.577 
  0.093   0.056  17.328 
```

it's also equivalent perf-wise (for now), but I feel this is no longer as clean using data frames, it moved from `sum(runif(2, min = -1, max = 1) ^ 2) < 1)` to `rowSums(replicate(2, runif(nrow(d), min=-1, max=1)) ^ 2) < 1)`. Any thoughts on how to simplify this closure?

```{r}
system.time({
sdf_len(sc, 10000) %>%
  spark_apply(function(d) rowSums(replicate(2, runif(nrow(d), min=-1, max=1)) ^ 2) < 1) %>%
  filter(id) %>% count() %>% collect() * 4 / 10000
})

user  system elapsed 
  0.090   0.055  17.520 
  0.090   0.055  17.472 
```

Finally, this issue is tracking performance improvements that need to be investigated: https://github.com/rstudio/sparklyr/issues/733. The first look at this point out that the initialization of R + Backend is most of the cost, this should not be as relevant for large jobs, but definitely significant for small jobs like the ones explored in this comment.

Merging...

kevinykuo:
@javierluraschi this breaks:

```
iris_tbl <- copy_to(sc, iris) %>%
  sdf_repartition(2L)

spark_apply(iris_tbl, function(data) {
  data[1:4] + rgamma(1,2)
})

 Error: org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 8.0 failed 1 times, most recent failure: Lost task 1.0 in stage 8.0 (TID 8, localhost, executor driver): java.lang.Exception: sparklyr worker rscript failure, check worker logs for details.
	at sparklyr.Rscript.init(rscript.scala:51)
	at sparklyr.WorkerRDD$$anon$2.run(rdd.scala:79)
```

javierluraschi:
@kevinykuo, interesting, will take a look after the merge with this one: https://github.com/rstudio/sparklyr/issues/737

catlain:
hi @javierluraschi  when i run example like below, it works:
```
spark_apply(iris_tbl, function(row) {
  row$Petal_Width + rgamma(1,2)
}, names = "randomized_petal_width")
```

but when i get a dataframe form a parquet file with schema below, it not works well:
```
root
 |-- mcc: integer (nullable = true)
 |-- mnc: integer (nullable = true)
```


i found error log like :
```
17/07/14 03:10:50 ERROR sparklyr: RScript (1712) terminated unexpectedly: list(message = "'names' attribute [1] must be the same length as the vector [0]", call = `colnames<-`(`*tmp*`, value = list("brand"))) 
17/07/14 03:10:50 INFO sparklyr: Worker (1712) completed wait using lock for RScript
17/07/14 03:10:50 INFO sparklyr: Gateway (1712) is terminating backend
```

at `take at NativeMethodAccessorImpl.java:0` task

catlain:
then i try:
```
spark_apply(test1, function(df) {
  data.frame(df$mcc + rgamma(1,2))
}, names = "mcc11")
```
seems there is no error in  stdout:
```
17/07/14 03:34:26 INFO sparklyr: Worker (3256) launching command Rscript --vanilla <source-file> 3256 FALSE;8891;localhost
17/07/14 03:34:26 INFO sparklyr: Worker (3256) is starting R process
17/07/14 03:34:26 INFO sparklyr: RScript (3256) is starting 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) is connecting to backend using port 8891 
17/07/14 03:34:26 INFO sparklyr: Gateway (3256) accepted connection
17/07/14 03:34:26 INFO sparklyr: Gateway (3256) is waiting for sparklyr client to connect to port 8891
17/07/14 03:34:26 INFO sparklyr: RScript (3256) is querying ports from backend using port 8891 
17/07/14 03:34:26 INFO sparklyr: Gateway (3256) received command 0
17/07/14 03:34:26 INFO sparklyr: Gateway (3256) found requested session matches current session
17/07/14 03:34:26 INFO sparklyr: Gateway (3256) is creating backend and allocating system resources
17/07/14 03:34:26 INFO sparklyr: Gateway (3256) created the backend
17/07/14 03:34:26 INFO sparklyr: Gateway (3256) is waiting for r process to end
17/07/14 03:34:26 INFO sparklyr: RScript (3256) found redirect gateway port 8891 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) is connected to backend 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) is connecting to backend session 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) is connected to backend session 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) created connection 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) is connected 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) retrieved worker context id 46 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) retrieved worker context 
17/07/14 03:34:26 INFO sparklyr: RScript (3256) found 5813 rows 
17/07/14 03:34:27 INFO sparklyr: RScript (3256) retrieved 5813 rows 
17/07/14 03:34:27 INFO sparklyr: RScript (3256) computing closure 
17/07/14 03:34:27 INFO sparklyr: RScript (3256) computed closure 
17/07/14 03:34:27 INFO sparklyr: RScript (3256) data.frame expected but list found 
17/07/14 03:34:28 ERROR sparklyr: RScript (3256) there is no package called ‘withr’NA 
17/07/14 03:34:28 INFO sparklyr: Worker (3256) completed wait using lock for RScript
17/07/14 03:34:28 INFO sparklyr: Gateway (3256) is terminating backend
```
but error in stderr:
```
17/07/14 03:34:28 INFO Executor: Executor is trying to kill task 0.9 in stage 38.0 (TID 379)
17/07/14 03:34:28 ERROR sparklyr: Gateway (3256) failed calling setResultArraySeq on 46
17/07/14 03:34:28 ERROR sparklyr: Worker (3256) failed to complete R process
(17/07/14 03:34:28 ERROR sparklyr: Worker (3256) failed to run rscript: ,java.lang.Exception: sparklyr worker rscript failure, check worker logs for details.)
17/07/14 03:34:28 WARN BlockManager: Putting block rdd_155_0 failed due to an exception
17/07/14 03:34:28 WARN BlockManager: Block rdd_155_0 could not be removed as it was not found on disk or in memory
(17/07/14 03:34:28 ERROR sparklyr: Gateway (3256) is shutting down with exception ,java.net.SocketException: Socket closed)
17/07/14 03:34:28 ERROR Executor: Exception in task 0.9 in stage 38.0 (TID 379)
java.lang.Exception: sparklyr worker rscript failure, check worker logs for details.
	at sparklyr.Rscript.init(rscript.scala:55)
	at sparklyr.WorkerRDD$$anon$2.run(rdd.scala:82)
```

PR #740, opened 2017-06-07
update README to use Spark 2.1

kevinykuo:
Keeping up with the times. Also `rsparkling` now supports more than just 1.6.


javierluraschi:
Thanks! This was a bit overdue!

PR #745, opened 2017-06-10
Forward port v0.5.6 fixes and use dplyr from CRAN.

javierluraschi:
Forward porting a couple new tests and the `top_n` fix to from https://github.com/rstudio/sparklyr/releases/tag/v0.5.6 and removing `dplyr` devel dependency and workarounds in travis.


javierluraschi:
Travis does not seem to pick on `dbplyr 1.0` probably CRAN mirror not fully propagated, will retry in a day or so.

Issue #679, opened 2017-05-15
unable to open spark log


dsblr:
Hello Sparklyr team,
I am unable to read spark log, I am using spark 2.0.1 and hadoop 2.6 version
Please help me to resolve the issue.
> spark_log(sc)
Error in open.connection(con, "rb") : cannot open the connection
In addition: Warning message:
In open.connection(con, "rb") :
  cannot open file 'log4j.spark.log': No such file or directory

BruceZhaoR:
same issue when used the latest `sparklyr`.  I encounter the issue #295 before, so I used `devtools` to install the latest `sparklyr`, here is the full warning and errors output:

```java
Error in open.connection(con, "rb") : cannot open the connection
In addition: Warning message:
In open.connection(con, "rb") :
  cannot open file 'log4j.spark.log': No such file or directory
Error: java.lang.IllegalArgumentException: Error while instantiating 'org.apache.spark.sql.hive.HiveSessionState':
	at org.apache.spark.sql.SparkSession$.org$apache$spark$sql$SparkSession$$reflect(SparkSession.scala:981)
	at org.apache.spark.sql.SparkSession.sessionState$lzycompute(SparkSession.scala:110)
	at org.apache.spark.sql.SparkSession.sessionState(SparkSession.scala:109)
	at org.apache.spark.sql.SparkSession$Builder$$anonfun$getOrCreate$5.apply(SparkSession.scala:878)
	at org.apache.spark.sql.SparkSession$Builder$$anonfun$getOrCreate$5.apply(SparkSession.scala:878)
	at scala.collection.mutable.HashMap$$anonfun$foreach$1.apply(HashMap.scala:99)
	at scala.collection.mutable.HashMap$$anonfun$foreach$1.apply(HashMap.scala:99)
	at scala.collection.mutable.HashTable$class.foreachEntry(HashTable.scala:230)
	at scala.collection.mutable.HashMap.foreachEntry(HashMap.scala:40)
	at scala.collection.mutable.HashMap.foreach(HashMap.scala:99)
	at org.apache.spark.sql.SparkSession$Builder.getOrCreate(SparkSession.scala:878)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

Issue #497, opened 2017-02-16
sparklyr error may be due to tmp/hive settings


dsblr:
Hello Sparklyr team,
I would like to draw your attention for the error I am getting. Because this is the error most of the people are getting when they work behind the corporate fire wall.

Error: 

library(sparklyr)
sc <- spark_connect(master = "local",spark_home = "C:\\Users\\user1\\AppData\\Local\\rstudio\\spark\\Cache\\spark-2.0.1-bin-hadoop2.4\\",hadoop_version = "2.4")
t1 <- copy_to(sc,iris)

## no error

but if I change the working directory and execute the same command I am getting the following error.

Error: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:422)
	at org.apache.spark.sql.hive.client.IsolatedClientLoader.createClient(IsolatedClientLoader.scala:258)
	at org.apache.spark.sql.hive.HiveUtils$.newClientForMetadata(HiveUtils.scala:359)
	at org.apache.spark.sql.hive.HiveUtils$.newClientForMetadata(HiveUtils.scala:263)
	at org.apache.spark.sql.hive.HiveSharedState.metadataHive$lzycompute(HiveSharedState.scala:39)
	at org.apache.spark.sql.hive.HiveSharedState.metadataHive(HiveSharedState.scala:38)
	at org.apache.spark.sql.hive.HiveSharedState.externalCatalog$lzycompute(HiveSharedState.scala:46)
	at org.apache.spark.sql.hive.HiveSharedState.externalCatalog(HiveSharedState.scala:45)
	at org.apache.spark.sql.hive.HiveSessionState.catalog$lzycompute(HiveSessionState.scala:50)
	at org.apache.spark.sql.hive.HiveSessionState.catalog(HiveSessionState.scala:48)
	at org.apache.spark.sql.hive.HiveSessionState$$anon$1.<init>(HiveSessionState.scala:63)
	at org.apache.spark.sql.hive.HiveSessionState.analyzer$lzycompute(HiveSessionState.scala:63)
[...]

Issue #466, opened 2017-02-01
/tmp/hive on HDFS should be writable error


dsblr:
Hi Team,
I am getting the following error while trying to copy data. Please help to resolve the issue.

Error: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:422)
	at org.apache.spark.sql.hive.client.IsolatedClientLoader.createClient(IsolatedClientLoader.scala:258)
	at org.apache.spark.sql.hive.HiveUtils$.newClientForMetadata(HiveUtils.scala:359)
	at org.apache.spark.sql.hive.HiveUtils$.newClientForMetadata(HiveUtils.scala:263)
	at org.apache.spark.sql.hive.HiveSharedState.metadataHive$lzycompute(HiveSharedState.scala:39)
	at org.apache.spark.sql.hive.HiveSharedState.metadataHive(HiveSharedState.scala:38)
	at org.apache.spark.sql.hive.HiveSharedState.externalCatalog$lzycompute(HiveSharedState.scala:46)
	at org.apache.spark.sql.hive.HiveSharedState.externalCatalog(HiveSharedState.scala:45)
	at org.apache.spark.sql.hive.HiveSessionState.catalog$lzycompute(HiveSessionState.scala:50)
	at org.apache.spark.sql.hive.HiveSessionState.catalog(HiveSessionState.scala:48)
	at org.apache.spark.sql.hive.HiveSessionState$$anon$1.<init>(HiveSessionState.scala:63)
	at org.apache.spark.sql.hive.HiveSessionState.analyzer$lzycompute(HiveSessionState.scala:63)
	at org.apache.spark.sql.hive.HiveSessionState.analyzer(HiveSessionState.scala:62)
	at org.apache.spark.sql.execution.QueryExecution.assertAnalyzed(QueryExecution.scala:49)
	at org.apache.spark.sql.Dataset$.ofRows(Dataset.scala:64)
	at org.apache.spark.sql.SparkSession.sql(SparkSession.scala:582)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

Issue #456, opened 2017-01-31
Error Trying to create tables on spark using copy_to


MumoNobert:
Hello, I am trying to get started on the sparklyr package but getting an error when I try to copy a  dataframe to a spark table/dataframe. I already have a spark connection.

library(sparklyr)
sc <- spark_connect(master = "local",spark_home="C:/Users/WayoPC/Documents/spark-2.0.2-bin-hadoop2.7")
mtcars_tbl <- copy_to(sc, mtcars, overwrite = TRUE)

ERROR
Error: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(Unknown Source)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(Unknown Source)
	at java.lang.reflect.Constructor.newInstance(Unknown Source)
	at org.apache.spark.sql.hive.client.IsolatedClientLoader.createClient(IsolatedClientLoader.scala:258)
	at org.apache.spark.sql.hive.HiveUtils$.newClientForMetadata(HiveUtils.scala:359)
	at org.apache.spark.sql.hive.HiveUtils$.newClientForMetadata(HiveUtils.scala:263)
	at org.apache.spark.sql.hive.HiveSharedState.metadataHive$lzycompute(HiveSharedState.scala:39)
	at org.apache.spark.sql.hive.HiveSharedState.metadataHive(HiveSharedState.scala:38)
	at org.apache.spark.sql.hive.HiveSharedState.externalCatalog$lzycompute(HiveSharedState.scala:46)
	at org.apache.spark.sql.hive.HiveSharedState.externalCatalog(HiveSharedState.scala:45)
	at org.apache.spark.sql.hive.HiveSessionState.catalog$lzycompute(HiveSessionState.scala:50)
	at org.apache.spark.sql.hive.HiveSessionState.catalog(HiveSessionState.scala:48)
	at org.apache.spark.sql.hive.HiveSessionState$$anon$1.<init>(HiveSessionState.scala:63)
	at org.apache.spark.sql.hive.HiveSessionState.analyzer$lzycompute(HiveSessionState.scala:63)
	at org.apache.spark.sql.hive.HiveSessionState.analyzer(HiveSessionState.scala:62)
	at org.apache.spark.sql.execution.QueryExecution.assertAnalyzed(QueryExecution.scala:49)
[...]

PR #755, opened 2017-06-15
Fix windows connection issue with newinstance0 present in logs

javierluraschi:
Several users have reported a `newinstance0 ` connection issue under Windows. See #98 #229 #277 #295 #354 #397 #409 #456 #466 #497 #633 #679.

I was able to repro this first time `sparklyr` is used in a new environment or by removing `HADOOP_HOME` and relaunching the RStudio.

This issue was being caused by using `SETX` to set the `HADOOP_HOME`. While this would work the next time the rsession launches (and the same session in some other environments); some other environments and packages, say `plumber`, `Conda`, etc. might not respect the `SETX` setting across rsession and trigger this error consistently.

The fix is to set `HADOOP_HOME` during `spark-submit` being specified through `withr::with_envvar`.

There is some risk of regressing scenarios where `SETX` was actually working appropriately. However, given that this has affected multiple Windows users, I think would be best to simplify this and perform an additional change to bring back `SETX` if we find environments where the new approach does not work correctly.

For those users affected by this error, please retry with `sparklyr 0.5.6-9002` or newer. You can install using `devtools::install_github("rstudio/sparklyr")`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
-Version: 0.5.6-9001
+Version: 0.5.6-9002
 Authors@R: c(
     person("Javier", "Luraschi", email = "javier@rstudio.com", role = c("aut", "cre")),
     person("Kevin", "Ushey", role = "aut", email = "kevin@rstudio.com"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -166,6 +166,9 @@
 
 ### Bug Fixes
 
+- Fixed `spark_connect` under Windows issue when `newInstance0` is present in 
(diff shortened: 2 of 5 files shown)


ldanai:
Hello,

I still got the same error using sparklyr (version: 0.8.4.9001). 

I installed using devtools::install_github("rstudio/sparklyr")

Appreciated if any suggestion

```
> sc <- spark_connect(master  = "spark://192.xxx.x.xx:8080", version = '2.1')
Error in file(con, "r") : cannot open the connection
In addition: Warning message:
In file(con, "r") :
  cannot open file 'C:\Users\xxxxx\AppData\Local\Temp\Rtmp4Q07Wi\file3c5c3a15674c_spark.log': Permission denied
```

PR #763, opened 2017-06-16
hive operators for regex matching

### LINKED ISSUE #329: Including grep, mataches, etc in mutate
Guys, as the other thread for udf and dapply keeping going i wonder if we could try to implement some of the reg. expression into sparklyr. 

Ex. df %>% mutate( text1 = ifelse( col1 %like% "$ text $" , 1, 0 )  ) - this works fine but if the reg expression has multiple attributes and multiple options the like ifelse becomes nasty. 

Ex: df %>% mutate( text1 = ifelse( col1 %like% c("$ text $","$ text2 $")  , 1, 0 )  )  - doesnt work. also is case sensitive. also way less powerful than grep. 

I can create a function that parse the sql element and takes care of all of this but it won't be aligned with grep and or other dplyr syntax. So before i do that, can you let me know if this is something you can easily integrate?

### LINKED ISSUE #761: filter Spark dataframe based on simple regex condition?
Hello,

I am struggling with a simple operation. I have a huge, gigantic csv file that I load in Spark using `spark_read_csv`

```
> df
# Source:   lazy query [?? x 1]
# Database: spark_connection
         Date
         <chr>
 1  hello my friend
 2  hello my dog
 3  hello my cat
```

and here I want to keep the rows in the **Spark dataframe** (no `collect()` allowed!!) that contains the word "cat". 

So this is a simple filter based on a basic regex condition. I have not been able to do so. I see a nice regex tokenizer available in `sparklyr` since 0.5, but how can I do the simple filtering above?

Thanks again!!

kevinykuo:
This change adds `%like%`, `%rlike%`, and `%regexp%` which correspond to `LIKE`, `RLIKE`, and `REGEXP`, respectively, from https://cwiki.apache.org/confluence/display/Hive/LanguageManual+UDF for regex matching.

Closes #761, closes #329

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -105,7 +105,7 @@
 ### Broom
   
 - Implemented `tidy()`, `augment()`, and `glance()` from tidyverse/broom for
-  `ml_model_generalized_linear_regression()` and `ml_model_linear_regression()`
+  `ml_model_generalized_linear_regression` and `ml_model_linear_regression`
   models.
   
 ### R Compatibility
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,9 @@
 
 ### dplyr
 
+- Support for Hive built-in operators `%like%`, `%rlike%`, and
+  `%regexp%` for matching regular expressions in `filter()` and `mutate()`.
+
 - Support for dplyr (>= 0.6) which among many improvements, increases
   performance in some queries by making use of a new query optimizer.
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -105,7 +108,7 @@
 ### Broom
   
 - Implemented `tidy()`, `augment()`, and `glance()` from tidyverse/broom for
-  `ml_model_generalized_linear_regression()` and `ml_model_linear_regression()`
+  `ml_model_generalized_linear_regression` and `ml_model_linear_regression`
```
kevinushey: Can you restore the parentheses here, please? (it's preferred to use those to make it more apparent that we're talking about functions as opposed to other kinds of objects)

kevinushey:
LGTM other than a small nitpick in the NEWS. Thanks for the PR!

NEWS.md
kevinykuo: Isn't `ml_model_generalized_linear_regression` the model object returned by the function `ml_generalized_linear_regression()`? Or can we also think of the model objects as functions somehow?

kevinushey: Ah, sorry about that -- you're exactly right! I misread the NEWS entry myself.

PR #766, opened 2017-06-19
Fix spark_connect(master = "local") when version isn't set

kevinykuo:
This PR fixes an issue introduced by 84e7ce09a1769440bb92bb7196c08db18c943158 where connection in local mode fails when `version` isn't specified.

Current behavior:

```
> sc <- spark_connect(master = "local")
Error in shell_connection(master = master, spark_home = spark_home, app_name = app_name,  : 
                            Failed to connect to Spark (SPARK_HOME is not set). 
```

With patch:

```
> sc <- spark_connect(master = "local")
* Using Spark: 2.0.0
```


javierluraschi:
Yes, this makes sense. Thank you, Kevin.

PR #773, opened 2017-06-21
Explicit local path reference while using copy_to in yarn client to a…

### LINKED ISSUE #525: Sparklyr copy_to fails
I am using the Sparklyr library to read and write data from R to HDFS. Reading data works as expected but writing gives issues.

To be able to use the spark_write_csv function, I need to convert my R data.frames into Spark objects. I use the sparklyr sdf_copy_to function for this (also tried with copy_to). However, I always get errors

Code:
```
table1 <- sdf_copy_to(sc,dataframe,OVERWRITE=TRUE)
spark_write_csv(table1, "path")
```
Error:

> `    Error: org.apache.hadoop.mapred.InvalidInputException: Input path does not exist: hdfs://iacchadoopdev01.dap:8020/tmp/Rtmp2gpelH/spark_serialize_62547a7b0f9ad206fd384af04e585deb3a2636ca7b1f026943d4cc1d11c7759a.csv`

Did somebody encountered the same problem and knows how to solve this?

A possible reason might be that the sdf_copy_to function stores the data onto my linux /tmp folder while the write function is looking for data on the HDFS /tmp folder.

Thanks in Advance!

javierluraschi:
…void hdfs path and fix #525 

Yes, I think all we need is the `file://` prefix for https://github.com/rstudio/sparklyr/issues/525, need to test this in `yarn-client` before the merge....

### DIFF
--- a/R/data_copy.R
+++ b/R/data_copy.R
@@ -130,7 +130,7 @@ spark_serialize_csv_scala <- function(sc, df, columns, repartition) {
     "sparklyr.Utils",
     "createDataFrameFromCsv",
     spark_context(sc),
-    tempfile,
+    paste("file://", tempfile, sep = ""),
     columns,
     as.integer(if (repartition <= 0) 1 else repartition),
     separator$regexp


R/data_copy.R
```diff
@@ -130,7 +130,7 @@ spark_serialize_csv_scala <- function(sc, df, columns, repartition) {
     "sparklyr.Utils",
     "createDataFrameFromCsv",
     spark_context(sc),
-    tempfile,
+    paste("file://", tempfile, sep = ""),
```
kevinykuo: The Scala code [here](https://github.com/rstudio/sparklyr/blob/master/java/spark-1.5.2/utils.scala#L323) seems to be trimming the slashes, maybe we need to pass an `URI` on the Scala side or manipulate the string there?

javierluraschi: I agree. Actually, if the issue is being reported with `master="local"`, we might rather want to replace [csv_file ](https://github.com/rstudio/sparklyr/blob/master/R/data_copy.R#L154) with `csv_file_scala`, since currently `csv_file` makes use of of `spark_read_csv` which is probably trying to read from HDFS in some corner cases when our expectation was to read from local. Long term, we want to move away from these custom serializers and use Apache Arrow / Feather, that is going to be a great day. Going to close this PR and follow up once we know what environment is being affected, I think we need to know that at the very minimum.

SemanticBeeng: Where can I learn more about work to integrate with `Apache Arrow`?
Curious what use cases you have in plans.

SemanticBeeng: Javier, where can I learn more about work to integrate with `Apache Arrow`?
Curious what use cases you have in plans and mind.

Issue #780, opened 2017-06-23
Some Scala exceptions cause backend to not return a status and crash r session


javierluraschi:
Create the following scala object and compile with `sparklyr`

```scala
object WorkerContext2 {
  def setResultArraySeq3(resultParam: Array[Any]) : String = {
    resultParam.map(x => x.asInstanceOf[Array[Any]])

    ""
  }
}
```

Run:

```r
library(sparklyr)
library(dplyr)
sc <- spark_connect(master = "local", version = "2.1.0")

testlist <- list(structure(list(Hit = "True", Project = "Blue", Year = "2011", Rating = "4", Launch = "26 Jan 2012", ID = "19", Dept = "1, 2, 4"),
                           .Names = c("Hit","Project", "Year", "Rating", "Launch", "ID", "Dept")),
                 structure(list(Hit = "False", Error = "Record not found"), .Names = c("Hit","Error")),
                 structure(list(Hit = "True", Project = "Green", Year = "2004",  Rating = "8", Launch = "29 Feb 2004", ID = "183", Dept = "6, 8"), .Names = c("Hit", "Project", "Year", "Rating", "Launch", "ID", "Dept")))

testdf <- bind_rows(testlist)

r2 <- testdf[2]
d2 <- lapply(1:nrow(r2), function(i) as.list(r2[i,]))

invoke_static(sc, "sparklyr.WorkerContext2", "setResultArraySeq3", d2)
```

Notice:

```
Error in core_invoke_method(sc, static, object, method, ...) : 
  No status is returned. The sparklyr backend might have failed.
```

From logs:

```
17/06/23 09:10:54 ERROR sparklyr: Backend (6023) failed calling exceptionString on sparklyr.Utils
java.lang.OutOfMemoryError: Java heap space
	at sparklyr.Serializer$.readStringBytes(serializer.scala:90)
	at sparklyr.Serializer$.readString(serializer.scala:99)
	at sparklyr.Serializer$$anonfun$readStringArr$1.apply(serializer.scala:141)
	at sparklyr.Serializer$$anonfun$readStringArr$1.apply(serializer.scala:141)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.immutable.Range.foreach(Range.scala:160)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
[...]

Issue #774, opened 2017-06-21
spark_apply can't deal with <NA> well


catlain:
---

spark_apply can't deal with "NA" well

```r
library(sparklyr)
library(dplyr)
spark_version <- "2.1.0"
sc <- spark_connect(master = "local", version = spark_version)

testlist <- list(structure(list(Hit = "True", Project = "Blue", Year = "2011", Rating = "4", Launch = "26 Jan 2012", ID = "19", Dept = "1, 2, 4"), 
                           .Names = c("Hit","Project", "Year", "Rating", "Launch", "ID", "Dept")), 
                 structure(list(Hit = "False", Error = "Record not found"), .Names = c("Hit","Error")), 
                 structure(list(Hit = "True", Project = "Green", Year = "2004",  Rating = "8", Launch = "29 Feb 2004", ID = "183", Dept = "6, 8"), .Names = c("Hit", "Project", "Year", "Rating", "Launch", "ID", "Dept")))

testdf <- bind_rows(testlist)
testdf_tbl <- sdf_copy_to(sc, testdf, overwrite = TRUE)

testdf_tbl
# Source:   table<testdf> [?? x 8]
# Database: spark_connection
    Hit Project  Year Rating      Launch    ID    Dept            Error
  <chr>   <chr> <chr>  <chr>       <chr> <chr>   <chr>            <chr>
1  True    Blue  2011      4 26 Jan 2012    19 1, 2, 4             <NA>
2 False    <NA>  <NA>   <NA>        <NA>  <NA>    <NA> Record not found
3  True   Green  2004      8 29 Feb 2004   183    6, 8             <NA>
```
get 1st col
```r
spark_apply(testdf_tbl, function(data) {
  data[1]
})

# Source:   table<sparklyr_tmp_7dd7d9d3887> [?? x 1]
# Database: spark_connection
    Hit
  <chr>
1  True
2 False
3  True
```
get 2nd col 
```r
spark_apply(testdf_tbl, function(data) {
  data[2]
})

ERROR: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 92.0 failed 1 times, most recent failure: Lost task 0.0 in stage 92.0 (TID 94, localhost, executor driver): java.lang.Exception: sparklyr worker rscript failure, check worker logs for details.
	at sparklyr.Rscript.init(rscript.scala:51)
	at sparklyr.WorkerRDD$$anon$2.run(rdd.scala:79)

Driver stacktrace:
[...]

PR #781, opened 2017-06-23
Avoid crashing r-session while invoking array character with NAs #780

javierluraschi:
The following crashes the r-session after executing a couple times the `invoke_static`, more info under https://github.com/rstudio/sparklyr/issues/780:

```r
library(sparklyr)
sc <- spark_connect(master = "local", version = "2.1.0")

naChar <- NA
class(naChar) <- "character"

invoke_static(
    sc,
    "sparklyr.Utils",
    "classExists",
    list(list(naChar), list("a")))
```

The problem here is that the function parameters with arrays of `character` types containing `NAs`, get serialized as an array of `characters` (correct) but when the `NA` gets serialized serialized to `void()` and nothing gets written in the binary buffer. This is fine while writing single `NA` in `character`s since the null type precedes, in case of arrays all have the same type. Since NAs are not supported in Scala character arrays.

On the scala side, since nothing is serialized for `NA`, the length of the character string is read afterward and allocates memory. In this case, it is too much memory that crashes the Scala backend with an OutOfMemory exception, which prevents the backend to writing the result back into the socket, which crashes the r-session when trying to read a value that is not in the socket buffer. This last part seems like a bug to me in R, taking a look at this as well since I think there are more ways to crash R following this pattern even if we fix this code path.


javierluraschi:
Found a couple more mappings related to https://github.com/rstudio/sparklyr/issues/774, introduced `spark_apply(data, function, debug = TRUE)` to allow us to attach as workers from a different r session and debug `spark_apply` calls.

Issue #779, opened 2017-06-22
 dplyr CUMPROD not available


gbisschoff:
Hi,

I want to run a cumprod in spark, but I get the error: 

> Undefined function: 'CUMPROD'

A simple (yet useless) example:
```
sc <- spark_connect(master = "local", version = '2.0.2')
iris_tbl <- copy_to(sc, iris,memory = TRUE,overwrite = TRUE)
iris_tbl%>%mutate(Roll_Prod=cumprod(Sepal_Length))
```

Real world application would be to calculate survival curves, yield curves etc.

Is there a way around this in the meantime?

Some session info:
```
R version 3.4.0 (2017-04-21)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

locale:
[1] LC_COLLATE=English_South Africa.1252  LC_CTYPE=English_South Africa.1252    LC_MONETARY=English_South Africa.1252
[4] LC_NUMERIC=C                          LC_TIME=English_South Africa.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] bindrcpp_0.1   dplyr_0.7.0    sparklyr_0.5.6

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.11      bindr_0.1         compiler_3.4.0    cellranger_1.1.0  dbplyr_1.0.0      base64enc_0.1-3   tools_3.4.0       digest_0.6.12    
 [9] jsonlite_1.4      tibble_1.3.3      rlang_0.1.1       openxlsx_4.0.17   shiny_1.0.3       DBI_0.6-1         rstudioapi_0.6    curl_2.6         
[17] yaml_2.1.14       parallel_3.4.0    haven_1.0.0       rio_0.5.5         withr_1.0.2       httr_1.2.1        hms_0.3           rappdirs_0.3.1   
[25] rprojroot_1.2     glue_1.0.0        data.table_1.10.4 R6_2.2.1          readxl_1.0.0      foreign_0.8-67    magrittr_1.5      feather_0.3.1    
[33] backports_1.1.0   htmltools_0.3.6   assertthat_0.2.0  mime_0.5          xtable_1.8-2      httpuv_1.3.3      config_0.2       
> 
```

javierluraschi:
@edgararuiz I wonder if you have already explored `CUMPROD` as part of your `dbplyr` work....

[...]

PR #789, opened 2017-06-27
Implement `cumprod()` for dplyr operations

javierluraschi:
Implements @kevinykuo suggestion to support for dplyr's `cumprod` using `exp` and `log`, see https://github.com/rstudio/sparklyr/issues/779.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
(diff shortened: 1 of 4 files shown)


randomgambit:
@javierluraschi @kevinushey @kevinykuo just to be sure: you are aware that EXP (sum_i log Xi) is going to **fail** if one of the Xi is negative, right? Also, you should use the natural log here.

Thanks!

PR #791, opened 2017-06-27
Use protocol to avoid path does not exist exception and fix #525

### LINKED ISSUE #525: Sparklyr copy_to fails

javierluraschi:
Fix for https://github.com/rstudio/sparklyr/issues/525 which was causing  `Path does not exist` exception during `copy_to` under systems configured with `HADOOP_HOME`.

Big thanks to @Wfore0990 for providing a docker image to repro this!

The problem is that, in some configurations, Spark adopts the `hdfs://` protocol as the default, fix is to explicitly call out the local file system.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
-Version: 0.5.6-9007
+Version: 0.5.6-9008
 Authors@R: c(
     person("Javier", "Luraschi", email = "javier@rstudio.com", role = c("aut", "cre")),
     person("Kevin", "Ushey", role = "aut", email = "kevin@rstudio.com"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -173,6 +173,9 @@
 
 ### Bug Fixes
 
+- Fixed `Path does not exist` referencing `hdfs` exception during `copy_to` under
+  systems configured with `HADOOP_HOME`.
+
 - Fixed session crash after "No status is returned" error by terminating
   invalid connection and added support to print log trace during this error.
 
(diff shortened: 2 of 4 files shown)


javierluraschi:
Unrelated, but found through repro, fixing regression on `spark_connect` with `version = NULL` but `SPARK_HOME` set.

The original code, for local mode, was meant to never find installed versions when `SPARK_HOME` was set. This was fine until some users find themselves with installed `SPARK_HOME` versions and setting the spark version but we were silently still using `SPARK_HOME`. Fix was made but broke `spark_connect` with `version=NULL`. Fix was made but broke `SPARK_HOME` set with `verson=NULL`. This one attempts the following:

- `SPARK_HOME=""` and `version=NULL`: Should install default version by looking for installed versions.
` `SPARK_HOME=""` and `version=something`: Same as above.
- `SPARK_HOME="path"` and `version=NULL`: Use `SPARK_HOME`
- `SPARK_HOME="path"` and `version=something`: Version overrides `SPARK_HOME`

PR #803, opened 2017-06-29
support weights.column in GLM type models

### LINKED ISSUE #217: ml_linear_regression regression weights
Can we add a regression weights parameter to `ml_linear_regression`, similar to the `weights` parameter in `lm()` in the base R package?

kevinykuo:
Closes #217

### DIFF
--- a/R/ml_generalized_linear_regression.R
+++ b/R/ml_generalized_linear_regression.R
@@ -28,7 +28,7 @@ ml_generalized_linear_regression <-
            features,
            intercept = TRUE,
            family = gaussian(link = "identity"),
-           weights_column = NULL,
+           weights.column = NULL,
            iter.max = 100L,
            ml.options = ml_options(),
            ...)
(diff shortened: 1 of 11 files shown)


R/ml_generalized_linear_regression.R
```diff
@@ -27,6 +28,7 @@ ml_generalized_linear_regression <-
            features,
            intercept = TRUE,
            family = gaussian(link = "identity"),
+           weights_column = NULL,
```
kevinushey: Given that other parameters here use `dot.case` rather than `snake_case`, should this be `weights.column`?

R/ml_generalized_linear_regression.R
```diff
@@ -88,6 +91,10 @@ ml_generalized_linear_regression <-
     invoke("setFamily", family) %>%
     invoke("setLink", link)
 
+  if (!is.null(weights_column))
+    model <- model %>%
+    invoke("setWeightCol", weights_column)
```
kevinushey: The indentation here is a bit misleading -- can we just wrap the expression in braces and indent continuations for the pipe operator?

kevinushey:
LGTM aside from some nitpicky stylistic concerns.

kevinykuo:
@kevinushey addressed comments; also corrected the regularization parameter in tests.

Could we have a feature request in the IDE to indent pipe continuations in `if` blocks with a single statement? :P

PR #816, opened 2017-07-05
Add support for columns parameter in spark_read_* functions to fix #799

### LINKED ISSUE #799: spark_read_json prohibitively slow with many json files
@kevinushey @kevinykuo I have about 40 very large json files (each is about 100GB) and I try to load them using

`spark_read_json(mypath/files_json*, memory = FALSE)`

However, **no matter what my cluster allocation is** (number of executors, number of cores, allocated RAM), the reading spawns 500k tasks and takes forever. I have the impression only one executor is trying to map all the data while this is something that can easily be parallelized. 

Some files can fit into the RAM of one cluster, but not all of them (some of them are much larger than the RAM available on a single node)

Any idea what `read_spark_json` is doing here?

javierluraschi:
Similar to `spark_read_csv()`, add support for `columns` on the other `spark_read_*()` functions, for instance:

```r
spark_read_json(
    sc,name = "iris_json_named",
    path = "test.json",
    columns = c("a", "b", "c")
  )
```

or

```r
spark_read_json(
    sc,name = "iris_json_typed",
    path = "test.json",
    columns = list("Sepal_Length" = "character", "Species" = "character", "Other" = "struct<a:integer,b:character>")
  )
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,9 @@
 
 ### External Data
 
+- Added `columns` parameter to `spark_read_*()` functions to load data with
+  named columns or explicit column types.
+
 - Added `partition_by` parameter to `spark_write_csv()`, `spark_write_json()`,
   `spark_write_table()` and `spark_write_parquet()`.
 
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -45,6 +45,7 @@ spark_csv_options <- function(header,
 #'   generated table. Use 0 (the default) to avoid partitioning.
 #' @param overwrite Boolean; overwrite the table with the given name if it
 #'   already exists?
+#' @param ... Optional arguments; currently unused.
 #'
 #' @details You can read data from HDFS (\code{hdfs://}), S3 (\code{s3n://}),
 #'   as well as the local file system (\code{file://}).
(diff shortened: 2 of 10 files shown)


kevinykuo:
Haven't had a chance to think too deeply on this, but do we really want users to be writing e.g. `"Other" = "struct<a:integer,b:character>"`? Would a nested list or similar be more intuitive?

kevinykuo:
Thinking of something like this:

```
df <- spark_read_json(
  sc, name = "iris_json_typed",
  path = "test.json",
  columns = list("Sepal_Length" = "character", 
                 "Species" = "character", 
                 "Other" = list("a" = "integer", "b" = "character"))
)
```

kevinykuo:
Also, these things can be nested a few levels deep (common in IoT data) so we want a straightforward way to specify the schema.

javierluraschi:
I don't disagree, at all. I stopped making changes when I realize that for nested fields, we also need `sdf_separate_columns()` that works for `StructFields`, otherwise, it's tricky to use the data imported with the example above.

I think there is certainly more work in this space, my intention here was to at least unblock loading tabular data by specifying the schema and see if performance is improved, if the nested field is the next blocking issue, we can send another PR.

PR #817, opened 2017-07-06
Integrate with spark-install package

javierluraschi:
Integrate https://github.com/rstudio/spark-install refactoring back to `sparklyr`.


javierluraschi:
@kevinushey @jjallaire this effectively moves the installation and windows configuration to a new `sparkinstall` package. We could explore other ways to take the scripts and avoid splitting `sparklyr` but this is probably the best approach. Notice that to the end user, everything is re-exported from `sparklyr` so they shouldn't even know about `sparkinstall`, we will also keep control over the function signatures in case we need to revert this work. 👍 👎 ?

tests/testthat/test-install-spark.R
```diff
@@ -4,7 +4,7 @@ test_that("supported spark_versions can be downloaded", {
   skip_on_cran()
   test_requires("RCurl")
 
-  versions <- spark_versions(latest = FALSE)
+  versions <- sparkinstall:::spark_versions(latest = FALSE)
```
kevinushey: Does this function need to be exported? (We shouldn't use `:::` when referring to functions from other packages)

R/install_spark_versions.R
```diff
@@ -1,149 +0,0 @@
-.globals <- new.env(parent = emptyenv())
```
kevinushey: Isn't this `.globals` object also used for extension packages (e.g. https://github.com/rstudio/sparklyr/blob/06d0a96fedecefd76d7a94c2a3ee47d6bf70baf4/R/spark_extensions.R#L15-L24)?

I think this needs to be re-defined in a separate file (perhaps in a new file called `aaa.R`, named as such just so it's sourced first by R)

kevinushey:
I was a bit out of the loop here, but what was the motivation for moving the installation code to a separate package? I'm always wary of splitting packages into multiple sub-packages since it inevitably leads to configuration headaches for users.

tests/testthat/test-install-spark.R
javierluraschi: I think we need to over this test to `sparkinstall` since it tests now a private function. I'll do that.

R/install_spark_versions.R
javierluraschi: Looks like we already did that but forgot to remove this instance, so this code was effectively duplicated, see: https://github.com/rstudio/sparklyr/blob/c436791ea967e2ad0fbdab36009479bf864fe0af/R/spark_globals.R

javierluraschi:
@kevinushey right, so here is how it went... Microsoft Azure folks reached out to see if we can work together mainly to:
1) Reuse installation paths, installable versions and windows configuration: Say, you don't want to install Spark multiple times under `AppData/Local/RStudio/spark` and `AppData/Local/Microsoft/spark` which means that versions and configuration also need to be aligned for this path to be sharable. This work is complete now.
2) Work together to compile and distribute `WINUTILS.exe` ourselves. This work has not started.

As a way to align the Python and R scripts, I created a new repo where we've been iterating on what we want the paths and versions file to be in Python and R. Since it's easier for me to work in a project, I created an R project for my own consumption, all good.

Now, we are at the point where we have the Python and R scripts ready (at least for phase-1) and we want to integrate the changes back to `sparklyr`. How we integrate the changes back has not been discussed yet; therefore, this PR.

A few approaches:
a) We manually pick the `sparkinstall` changes into `sparklyr`.
b) We automatically pick the `sparkinstall` changes into `sparklyr`.
c) We take a dependency on `sparkinstall` through CRAN.

I would preffer an automated approach, so either (b) or (c) work for me. For (b), seems that this would have to be a custom process that I have not seen other packages using, so in my mind, seems we have to go for (c), but 'm hoping we can use this PR to discuss how to best support this.

kevinushey:
Although I'm probably saying this too late, it _seems_ like we should only be maintaining + synchronizing a set of configuration files (e.g. the 'common' files here: https://github.com/rstudio/spark-install/tree/master/common), and we should be manually pulling those down into `sparklyr` at regular intervals, and then using those for installation.

Unless Microsoft / other environments want a package separate from `sparklyr` specifically for installation of Spark?

I would lean towards 2) as well. Assuming the R code within `sparkinstall` is self-contained, and we do want to keep the Spark installation code within a separate package as well, it could potentially be as simple as just copying the R files from `sparkinstall` to `sparklyr` as needed.

jjallaire:
Is the idea that we are developing the code in sparkinstall just so that
changes can by synced between R and Python? That seems sensible, but I
agree that having another package is likely to be more trouble than it's
worth. I'd just write a script that imports the sparkinstall R code into
sparklyr on demand.

javierluraschi:
@kevinushey they don't require a separate package. I'm happy to give (2) a shot, thanks for the feedback JJ/Kevin.

Issue #625, opened 2017-04-17
Refactor databricks gateway launch into indendent codebase


javierluraschi:

javierluraschi:
Sent refactoring PR: https://github.com/rstudio/sparklyr/pull/820

PR #820, opened 2017-07-07
Fix sparklyr under databricks and refactor connections

javierluraschi:
1. Fix `sparklyr` under Databricks, the fix is to expose the `sparklyr.Backend.main` scala API that Databricks still uses, we will remove once Databricks releases.
2. Refactored Databricks connection code as a gateway connection. Already tested this in Databricks cluster. Closes: https://github.com/rstudio/sparklyr/issues/625

screen shot 2017-07-06 at 10 24 19 pm

CC: @falaki


javierluraschi:
Notice that `spark_apply` was broken under databricks. The problem was that databricks sets a different default gateway port, that when a worker starts assumes the role of gateway but was not terminating the thread correctly and was instead processing the `RegisterSession` gateway command but shutting down afterwards, all is good now:

screen shot 2017-07-10 at 5 03 36 pm

R/spark_apply.R
```diff
@@ -68,17 +68,25 @@ spark_apply <- function(x, f, names = colnames(x), memory = TRUE, ...) {
...
+  worker_config <- worker_config_serialize(
+    c(
+      list(
+        debug = isTRUE(args$debug)
+      ),
+      sc$config
+    )
+  )
+
+  worker_port <- spark_config_value(sc$config, "sparklyr.gateway.port", "8880")
```
falaki: Is this port number identical with driver gateway port?

R/spark_gateway.R
```diff
@@ -34,7 +35,7 @@ gateway_connection <- function(master, config) {
 spark_gateway_connection <- function(master, config, gatewayInfo, gatewayAddress) {
   tryCatch({
     # set timeout for socket connection
-    timeout <- spark_config_value(config, "sparklyr.backend.timeout", 60)
+    timeout <- spark_config_value(config, "sparklyr.backend.timeout", 30 * 24 * 60 * 60)
```
falaki: Does such a long timeout make sense?

java/spark-1.5.2/backend.scala
```diff
@@ -276,6 +283,13 @@ class Backend {
                     logger.log("is terminating")
                     System.exit(0)
                   }
+
+                  // workers should always terminate but without exceptions
+                  if (isWorker) {
+                    logger.log("is terminating worker")
```
falaki: Maybe a more informative log message: `sparklyr worker backend terminating` or something like that?

R/spark_apply.R
javierluraschi: Yes, so far has worked fine. When there is only one driver, the main gateway serves all the mapping requests. For worker roles, they would start its own gateway.

R/spark_gateway.R
javierluraschi: Yes, actually this was a fix performed over `sparkR`. Someone mentioned at Spark Summit SFO that they weren't using `sparkR` since it would timeout constantly for really long jobs, not 100% sure this was the main timeout we fixed, but I'm almost certain. So if anything, you might want to port this back to `sparkR`.

java/spark-1.5.2/backend.scala
javierluraschi: `logger.log` concatenates mostly the info you are requesting, but currently, it outputs:

```
17/07/10 17:21:32 INFO sparklyr: Worker (4778) is terminating worker
```

To your point, I think we rather want the following:

```
17/07/10 17:21:32 INFO sparklyr: Worker (4778) is terminating backend
```

Fixed.

R/spark_apply.R
falaki: On DB I am using random port number for gateway. I guess since it worked in your test, we are fine.

PR #822, opened 2017-07-07
add csrf_header to livy_config()

kevinykuo:
This PR adds the `csrf_header` argument to `livy_config()` which allows users to connect to Livy servers with CSRF protection enabled.

In Livy, this function is disabled by default, but when it is enabled, we need to append a custom http-header.

Relevant bits from Livy:
https://github.com/cloudera/livy/blob/2abb8a3d2850c506ffd2b8a210813f1b8353045f/conf/livy.conf.template#L63-L65
https://github.com/cloudera/livy/blob/91058d29846a82b9d006390a6521bcba11bf56bd/server/src/main/scala/com/cloudera/livy/server/CsrfFilter.scala#L26-L49

I've tested this on an HDInsight cluster.

...As I'm writing this, I'm thinking maybe we should just hardcode this in `livy_config()` so users never have to worry about it??

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -153,6 +153,9 @@
 - While connecting to spark from Windows, setting the `sparklyr.verbose` option
   to `TRUE` prints detailed configuration steps.
   
+- Added `csrf_header` to `livy_config()` to enable connections to Livy servers with
+  CSRF protection enabled.
+  
 ### Compilation
 
 - Added support for `jar_dep` in the compilation specification to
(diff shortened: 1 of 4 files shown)


javierluraschi:
@kevinykuo I believe there shouldn't be any side effect from setting this header, if a system is not configured to use CSRF, it would be ignored. Your suggestion to make it the default does seem like a better approach.

Out of curiosity, could you share which Livy system you find out requires this?

kevinykuo:
@javierluraschi Livy >= 0.3 supports this feature although it's off by default. The sysadmin at work decided to turn this on and we're running Azure.

PR #824, opened 2017-07-10
Update utils.R

chris-ukds:
The java binary on windows is called `java.exe`, so the current code for `get_java()` produces the following error when initializing the Spark context:
```
> sc = spark_connect("local")
   Error in get_java(throws = TRUE) : 
   Java is required to connect to Spark. JAVA_HOME is set but does not point to a valid version. Please fix   
   JAVA_HOME or reinstall from: https://www.java.com/en/
```
FYI:
```
> sessionInfo()
R version 3.3.3 (2017-03-06)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 14393)
```


javierluraschi:
@chris-ukds thank you!

PR #834, opened 2017-07-12
Added support for custom headers in config

abhijitjaiswal:
Ability to add custom headers to rest call

### DIFF
--- a/R/livy_connection.R
+++ b/R/livy_connection.R
@@ -48,7 +48,7 @@ livy_validate_http_response <- function(message, req) {
 #'
 #' @return Named list with configuration data
 livy_config <- function(config = spark_config(), username = NULL, password = NULL,
-                        custom_headers=NULL, csrf_header = TRUE) {
+                        custom_headers = NULL, csrf_header = TRUE) {
   if (!is.null(username) || !is.null(password)) {
     secret <- base64_enc(paste(username, password, sep = ":"))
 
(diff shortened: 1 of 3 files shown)


kevinykuo:
@abhijitjaiswal Thanks for the PR! Could you describe the use case(s) for this?

abhijitjaiswal:
@kevinykuo The following is the use case:
Livy is running behind a web service that needs custom header to be passed for authentication before proxying request to Livy.

It is clean backward compatible change.

R/livy_connection.R
```diff
@@ -48,7 +48,7 @@ livy_validate_http_response <- function(message, req) {
 #'
 #' @return Named list with configuration data
 livy_config <- function(config = spark_config(), username = NULL, password = NULL,
-                        csrf_header = TRUE) {
+                        custom_headers=NULL, csrf_header = TRUE) {
```
kevinykuo: Could we have spaces on either side of `=`?

R/livy_connection.R
```diff
@@ -70,6 +70,12 @@ livy_config <- function(config = spark_config(), username = NULL, password = NUL
     )
   }
 
+  if(!is.null(custom_headers)){
```
kevinykuo: Could we have a space before each `{`?

kevinykuo:
LGTM. @abhijitjaiswal Would you mind also adding a entry to `NEWS` under "Connections"?

@javierluraschi I think for simplicity we may want to get rid of `csrf_header` and default `custom_headers` to `list("X-Requested-By" = "sparklyr")`. If the user modifies the default, it'll still work with CSRF protection since any custom header gets through Livy's filter. We can do this in a separate PR.

abhijitjaiswal:
@kevinykuo All suggestions are implemented, could you please merge the changes to master. Thanks for the quick responses 👍

javierluraschi:
@kevinykuo your suggestion makes sense, merging this one then.

PR #837, opened 2017-07-12
Fix compilation in sparklyr extensions by allowing them to not have versioned scala dirs

javierluraschi:
Currently `compile_package_jars()` under  https://github.com/javierluraschi/sparkhello and other extensions does not work since we introduced support for versioned scala directories. This change fixes compilation in `sparklyr` extensions by allowing them to not have versioned scala dirs.

CC: @kevinykuo

### DIFF
--- a/R/spark_compile.R
+++ b/R/spark_compile.R
@@ -234,36 +234,39 @@ find_jar <- function() {
 #' Spark extensions, when used with \code{\link{compile_package_jars}}.
 #'
 #' @param pkg The package containing Spark extensions to be compiled.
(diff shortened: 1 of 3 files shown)


kevinushey:
I think it would be better if we just always included top-level Scala sources (ie the `make_version_filter()` function should provide a filter that accepts top-level Scala sources as well)

kevinykuo:
I agree with @kevinushey, don't think we need to introduce additional arguments/complexity here.

javierluraschi:
I like that too, I'll make the change.

PR #858, opened 2017-07-22
Distribute Driver Packages to Worker Nodes for `spark_apply()`

javierluraschi:
`spark_apply` currently expects all packages to be pre-installed in each worker node; while this is feasible, it puts an unnecessary burden on system administrators.


javierluraschi:
Was able to run:

```r
sparkwarc::spark_read_warc(
    sc,
    "first_warc",
    sparkwarc::cc_warc(1,1),
    overwrite = TRUE,
    group = TRUE,
    include = "Content-Length")
```

Notice that `sparkwarc::spark_read_warc()` uses `spark_apply()` (see [sparkwarc.R#L63-L86](https://github.com/javierluraschi/sparkwarc/blob/feature/spark-apply-rcpp/R/sparkwarc.R#L63-L86))). Inside `spark_apply()`, it to download the ~1GB WARC file and reads this in about ~2min using `Rcpp`! This is about 10X faster than what I was able to process in my UseR demo earlier in the year, that said, I'm missing to run this at scale but looks promising.

Here is the output:

```
# Source:   table<first_warc> [?? x 2]
# Database: spark_connection
    tags                                              content
   <dbl>                                                <chr>
 1     1                            "Content-Length: 344\r\n"
 2     2                            "Content-Length: 244\r\n"
 3   492 "Content-Length: 15512\r\nContent-Length: 15299\r\n"
 4     3                             "Content-Length: 19\r\n"
 5     2                            "Content-Length: 214\r\n"
 6   424 "Content-Length: 18700\r\nContent-Length: 18396\r\n"
 7     3                             "Content-Length: 20\r\n"
 8     2                            "Content-Length: 213\r\n"
 9   642 "Content-Length: 30391\r\nContent-Length: 29927\r\n"
10     3                             "Content-Length: 20\r\n"
# ... with 1.66e+05 more rows
```

And counting all tags for this WARC:

```
tbl(sc, "first_warc") %>% select(tags) %>% summarize(total_tags = sum(tags))
# Source:   lazy query [?? x 1]
# Database: spark_connection
  total_tags
       <dbl>
1   90263846
```

javierluraschi:
## Using Scala:

```r
system.time(
  r1 <- sparkwarc::spark_read_warc(
    sc,
    "first_warc",
    sparkwarc::cc_warc(1,1),
    overwrite = TRUE,
    group = TRUE
    parse = TRUE
  )
)

    user   system  elapsed 
   3.132    4.260 1000.531 
```

## Using spark_apply() and Rcpp::

```r
system.time(
  r1 <- sparkwarc::spark_read_warc(
    sc,
    "first_warc",
    sparkwarc::cc_warc(1,1),
    overwrite = TRUE,
    group = TRUE,
    include = "Content-Length",
    filter = "movies"
  )
)

   user  system elapsed 
  0.560   0.392  97.151 
```

Issue #737, opened 2017-06-07
spark_apply not working with sdf_repartition


javierluraschi:
From @kevinykuo:

```
iris_tbl <- copy_to(sc, iris) %>%
  sdf_repartition(2L)

spark_apply(iris_tbl, function(data) {
  data[1:4] + rgamma(1,2)
})

 Error: org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 8.0 failed 1 times, most recent failure: Lost task 1.0 in stage 8.0 (TID 8, localhost, executor driver): java.lang.Exception: sparklyr worker rscript failure, check worker logs for details.
	at sparklyr.Rscript.init(rscript.scala:51)
	at sparklyr.WorkerRDD$$anon$2.run(rdd.scala:79)
```

kevinykuo:
This is now working for some reason (also works with `memory = FALSE`). Let's leave open until we have a test?

javierluraschi:
Looks fixed by https://github.com/rstudio/sparklyr/pull/883, can't repro this anymore.

Issue #878, opened 2017-07-30
Using `spark_apply()` with empty partitions fails


javierluraschi:
Processing empty partitions currently fails, see:

```r
sdf_len(sc, 10) %>%
  sdf_repartition(10) %>%
  spark_apply(function(e) e) %>% head(2)
```

```
17/07/30 11:04:19 INFO ShuffleBlockFetcherIterator: Started 0 remote fetches in 0 ms
17/07/30 11:04:19 INFO sparklyr: RScript (8762) retrieved 0 rows 
17/07/30 11:04:19 INFO sparklyr: RScript (3719) retrieved 0 rows 
17/07/30 11:04:19 INFO sparklyr: RScript (3460) retrieved 2 rows 
17/07/30 11:04:19 ERROR sparklyr: RScript (8762) terminated unexpectedly: list(message = "'names' attribute [1] must be the same length as the vector [0]", call = names(x) <- value) 
17/07/30 11:04:19 ERROR sparklyr: RScript (3719) terminated unexpectedly: list(message = "'names' attribute [1] must be the same length as the vector [0]", call = names(x) <- value) 
```

javierluraschi:
Fixed with https://github.com/rstudio/sparklyr/pull/883

PR #883, opened 2017-07-31
For empty partitions, `spark_apply()` fails.

javierluraschi:
See https://github.com/rstudio/sparklyr/issues/878. The following command fails:

```r
sdf_len(sc, 10, repartition = 12) %>% spark_apply(function(e) e)
```

Since out of the 10 partitions, at least 2 are empty. This might be more common while working at scale. Fix is to avoid retrieving data when no rows are provided in the partition.


kevinykuo:
I think we can also close https://github.com/rstudio/sparklyr/issues/737 with this PR

PR #890, opened 2017-08-01
Fix slow printing of spark dataframes

javierluraschi:
See https://github.com/tidyverse/tibble/issues/285.

Bottom line, `print.tbl_spark` is too slow since `dbplyr` started using `tibble` as the default way of printing records, but `tibble::truncate` uses `nrow()` which uses `dim` which resolves to `dim.tbl_spark` and cause a full run over the dataset.

I explored a few alternative solutions:
1. Implement `print.tbl_spark`: Not great since a lot of functionality is implemented as private methods in `dbplyr` so we would have to copy all that or remove funcitonality like printing `order-bys`.
2. Workaround in `tibble`, hard to coordinate this since this is high-pri for `sparklyr`.

Users that make use of `dim`, `nrow` and `ncol` would have to use `sdf_dim`, `sdf_nrow` and `sdf_ncol` instead.


javierluraschi:
+ @kevinushey

JohnMount:
I would argue that `tibble::truncate` is not the only user of `nrow()` in the world- so changing the behavior of `nrow()` just for it is not appropriate.   Often one wants to write analysis code that is generic (works without changing over local data, `dbplyr`, and `sparklyr`).  Most of my calls to `nrow()` are to check if a table is empty or not, so if there were a generic way to do that that would help a lot.  Also there must be some way to support `ncol()` efficiently.  And finally if you are going to do this it makes more sense to return `c(NA, NA)` for dim(), not `NA`.  Perhaps the right solution is a pull-request back to `tibble::truncate`.

PR #893, opened 2017-08-02
Fix typo in error message

martinstuder:
Changed yarn-size.xml to yarn-site.xml


javierluraschi:
@martinstuder thank you!

Issue #885, opened 2017-08-01
Reconcile arguments for hyperparameters in ml_* functions with Spark


kevinykuo:
...and make sure we're exposing all the available arguments.

kevinykuo:
RF https://github.com/rstudio/sparklyr/issues/377

Issue #377, opened 2016-12-12
How to control Random Forest


dsblr:
Hi Team,
this is not an issue, but this is a query to know basic functionality of **sparklyr ml_random_forest** package.
I would like to know how can I do the following operations ? 
1) to control the number of variable at each split
2) to control the sample size
3) to control the cutoff value

Please help me to do the above parameter tuning.

kevinykuo:
We should expose these arguments/document them so users don't have to pass them through `ml.options`. Will work on this.

kevinykuo:
@dsblr these have been implemented and documented as of https://github.com/rstudio/sparklyr/pull/896. Please give the latest dev version a try and feel free to open another issue if needed.

dsblr:
@kevinykuo Thanks for your support. Is there a way we can find OOB(out of bag error) from trained random forest model as R Random forest model. Please help me on OOB for sparklyr Random Forest.

PR #896, opened 2017-08-03
enhancements for ml_random_forest()

kevinykuo:
This PR adds the following tuning arguments to `ml_random_forest()`:

- `col.sample.rate`: Column sampling rate at each split (*not* per tree.) Spark ML calls this `featureSubsetStrategy` and requires a *string* input, e.g. "onethird", "sqrt", or "n", with `"when n is in the range (0, 1.0], use n * number of features. When n is in the range (1, number of features), use n features"`... which I think is really confusing. In this PR I'm requiring a *double* between 0 and 1. `NULL` defaults to the Spark ML `auto` strategy which equates to `sqrt(k)/k` for classification and `1/3` for regression, where `k` is number of features.

- `impurity`: The measure for info gain calculation at each split. Spark ML supports one of "gini" and "entropy" for classification (defaults to gini) and "variance" for regression. Input validation is included in our implementation.

- `min.info.gain`: Info gain required to split. This is an absolute number and not relative improvement.

- `min.rows`: Minimum number of observations each node must have after split. Spark ML calls this `minInstancesPerNode`, h2o calls this `min_rows` and R randomForest calls it `nodesize`. I like h2o's choice here.

- `thresholds`: This is used in multiclass classification to introduce bias to the predictions. Basically users can pass a vector of numbers (t_1, t_2, ..., t_K) where K is the number of classes, and the probabilities will be adjusted to be (p_1/t_1, p_2/t_2, ..., p_K/t_K), where the p_i's are the raw probabilities. The predicted label would be the one with highest adjusted probability.

- `seed`: Used for column and row sampling.

### DIFF
--- a/R/ml_random_forest.R
+++ b/R/ml_random_forest.R
@@ -5,9 +5,11 @@
 #' @template roxlate-ml-x
 #' @template roxlate-ml-response
 #' @template roxlate-ml-features
+#' @template roxlate-ml-decision-trees-col-sample-rate
 #' @template roxlate-ml-decision-trees-max-bins
 #' @template roxlate-ml-decision-trees-max-depth
 #' @template roxlate-ml-decision-trees-min-info-gain
+#' @template roxlate-ml-decision-trees-min-rows
 #' @template roxlate-ml-decision-trees-num-trees
 #' @template roxlate-ml-decision-trees-type
 #' @template roxlate-ml-options
@@ -19,9 +21,11 @@
 ml_random_forest <- function(x,
                              response,
                              features,
+                             col.sample.rate = NULL,
                              max.bins = 32L,
                              max.depth = 5L,
                              min.info.gain = 0,
+                             min.rows = 1L,
                              num.trees = 20L,
                              type = c("auto", "regression", "classification"),
                              ml.options = ml_options(),
@@ -43,9 +47,19 @@ ml_random_forest <- function(x,
     ml.options = ml.options
   )
 
+  col.sample.rate <- ensure_scalar_double(col.sample.rate, allow.null = TRUE)
+  if (!is.null(col.sample.rate)) {
+    if (!(col.sample.rate > 0 && col.sample.rate <= 1))
+      stop("'col.sample.rate' must be in (0, 1]")
+    col.sample.rate <- ensure_scalar_character(as.character(col.sample.rate))
+  } else {
+    col.sample.rate <- "auto"
+  }
+
   max.bins <- ensure_scalar_integer(max.bins)
   max.depth <- ensure_scalar_integer(max.depth)
   min.info.gain <- ensure_scalar_double(min.info.gain)
+  min.rows <- ensure_scalar_integer(min.rows)
   num.trees <- ensure_scalar_integer(num.trees)
   type <- match.arg(type)
   only.model <- ensure_scalar_boolean(ml.options$only.model)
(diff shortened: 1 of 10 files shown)


kevinykuo:
Related: https://github.com/rstudio/sparklyr/issues/885, https://github.com/rstudio/sparklyr/issues/377

javierluraschi:
@kevinykuo Looks like some of the parameters, say `setFeatureSubsetStrategy ` are not available in `1.6`: https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.ml.classification.RandomForestClassificationModel, throw a `stop("X parameter is unsupported in Spark ", spark_version(sc))` for `!is.null(param.name) && spark_version(sc) < "2.0"`?

kevinykuo:
@javierluraschi I remember everything I added was annotated `@Since("1.4.0")` in the Scala source... I'll take a look.

kevinykuo:
OK found the issue https://github.com/apache/spark/commit/da60b34d2f6eba19633e4f1b46504ce92cd6c179. So support for entering an actual number for the rate was added for `2.0`, which means `1.6` users would have to type things like `"onethird"` or `"log2"`. Let me think about how to deal with this...

Will also add tests to ensure compatibility and run `run-tests`.

javierluraschi:
@kevinykuo I think is fine to mention in the docs that, for Spark 1.6 valid values are `"onethird"`, `"log2"`, etc. but Spark 2.0 supports numeric values. I don't think we need to support numeric in 1.6, in general, we try not to enhance Spark too much but rather, to provide sensible wrappers.

javierluraschi:
@kevinykuo I'm thinking we can leave this one out of the `sparklyr 0.6.1` patch, correct? Let me know if you think otherwise. We can merge this PR when ready, since I would rather branch off the `sparklyr 0.6.1` release in a separate branch.

kevinykuo:
@javierluraschi yeah I didn't mean for this PR (and other upcoming ML enhancements) to be in the `sparklyr 0.6.1` patch.

R/ml_random_forest.R
```diff
@@ -41,11 +53,41 @@ ml_random_forest <- function(x,
...
+      if (col.sample.rate == 1)
+        col.sample.rate <- "all"
+      else {
+        # Prior to Spark 2.0.0, random forest does not support arbitrary
+        #   column sampling rates. So we map the input to one of the supported
+        #   strategies: "onethird", "sqrt", or "log2".
+        k <- length(features)
+        strategies <- dplyr::data_frame(strategy = c("onethird", "sqrt", "log2"),
+                                        rate = c(1/3, sqrt(k)/k, log2(k)/k)) %>%
+          arrange(!! rlang::sym("rate"))
```
kevinushey: This could probably just be `arrange(rate)`?

kevinykuo: Wouldn't R CMD check complain that rate isn't defined? Is there a way around that?

R/ml_random_forest.R
```diff
@@ -41,11 +53,41 @@ ml_random_forest <- function(x,
     ml.options = ml.options
   )
 
+  col.sample.rate <- ensure_scalar_double(col.sample.rate, allow.null = TRUE)
```
kevinushey: Should we allow users to specify strategies by name directly? (e.g. supply `"onethird"` explicitly)

kevinushey: Sorry; I see now this decision was made explicitly. I think it's fine to enforce that this has to be a numeric value.

R/ml_random_forest.R
kevinushey: Ahh, good point. One can resolve this using e.g. `utils::globalVariables()` but that's a pretty ugly hammer to wield here. Given that this is to avoid R CMD check issues I think this is fine.

kevinushey:
LGTM; @javierluraschi I'll let you merge when ready.

PR #932, opened 2017-08-14
Enhancements for tree-based methods

kevinykuo:
TODO (note to self):
- [x] Investigate implementation of RF to see why a 1-tree forest doesn't agree with a single decision tree like a 1-tree GBT ensemble.
- [x] Implement advanced parameter setters for `cacheNodeIds`, `maxMemoryInMB`, and `checkpointInterval`.
~~- [ ] More input checking on R side to minimize Java stack traces.~~
- [x] Consistent internal variable names for models.
- [x] Tests for error messages for features not supported in old versions.
~~- [ ] More coverage overall for these methods.~~

Related: https://github.com/rstudio/sparklyr/issues/885

### DIFF
--- a/R/ml_decision_tree.R
+++ b/R/ml_decision_tree.R
@@ -10,6 +10,7 @@
 #' @template roxlate-ml-decision-trees-max-depth
 #' @template roxlate-ml-decision-trees-min-info-gain
 #' @template roxlate-ml-decision-trees-min-rows
+#' @template roxlate-ml-decision-trees-thresholds
 #' @template roxlate-ml-decision-trees-type
 #' @template roxlate-ml-decision-trees-seed
 #' @template roxlate-ml-options
@@ -27,6 +28,7 @@ ml_decision_tree <- function(x,
                              min.info.gain = 0,
                              min.rows = 1L,
                              type = c("auto", "regression", "classification"),
+                             thresholds = NULL,
                              seed = NULL,
                              ml.options = ml_options(),
                              ...)
@@ -53,6 +55,7 @@ ml_decision_tree <- function(x,
   min.rows <- ensure_scalar_integer(min.rows)
   type <- match.arg(type)
   only.model <- ensure_scalar_boolean(ml.options$only.model)
+  thresholds <- if (!is.null(thresholds)) lapply(thresholds, ensure_scalar_double)
   seed <- ensure_scalar_integer(seed, allow.null = TRUE)
 
   envir <- new.env(parent = emptyenv())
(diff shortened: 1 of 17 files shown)


kevinykuo:
So it turns out that this line of [documentation in Spark](https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.ml.regression.RandomForestRegressor) is incorrect regarding `featureSubsetStrategy`: 

> "n": when n is in the range (0, 1.0], use n * number of features. When n is in the range (1, number of features), use n features. (default = "auto")

The issue is that `"1"` is being interpreted as *one* feature, and not *1.0 times number of features*. In other words, the interval in *When n is in the range (1, number of features), use n features* is actually closed. The fix is to call `format(., nsmall = 1L)` on `col.sample.rate`.

kevinykuo:
For GBT checkpointing with `checkpoint.interval = 1L` is failing with 
```
Error: java.io.FileNotFoundException: File file:/private/var/folders/1z/808_xcnx2fx8d9d2l4d8_hqr0000gn/T/RtmpORrZEf/5e668864-a8c5-4035-b312-afc850834c6c/rdd-3391/part-00000 does not exist
```
 but works fine with other interval parameters. Likely a Spark bug; won't investigate further for now since this should hit very few use cases if any.

kevinykuo:
Summary of changes in this PR:

- Finished implementing all tuning options for `ml_decision_tree()`, `ml_gradient_boosted_trees()`, and `ml_random_forest()`.
- Fixed a bug for `col.sample.rate` in `ml_random_forest()`.
- Cleaned up internal variable names a bit
- Added a few tests.

I've decided to limit the scope this PR here. While going through the code I think there's opportunity to refactor some of the input checking code, since these tree-based methods share many hyperparameters, but that's for another day.

R/ml_decision_tree.R
```diff
@@ -57,25 +79,48 @@ ml_decision_tree <- function(x,
...
-    classifier
-  else if (responseType %in% c("DoubleType", "IntegerType"))
-    regressor
-  else
-    classifier
-
-  rf <- invoke_new(sc, envir$model)
-
-  model <- rf %>%
+  modelType <- if (identical(type, "regression"))
```
kevinushey: Do you think it's worth moving this code (+ the code around `impurity`) to a set of helper functions?

man-roxygen/roxlate-ml-decision-trees-learn-rate.R
```diff
@@ -0,0 +1 @@
+#' @param learn.rate the learning rate or step size, defaults to 0.1.
```
kevinushey: nit: capital `The`?

R/ml_decision_tree.R
kevinykuo: Yep. I also think we should have helper functions for most of the common inputs so we're not copy-pasting a bunch of `ensure_scalar_`-`if-stop` code. Thought we'd do it in another PR though.

kevinushey:
LGTM! Only comment is whether you might consider refactoring some of the duplicated code into a single helper function, but I'm fine leaving it as-is as well.

R/ml_decision_tree.R
kevinushey: Sounds good!

Issue #194, opened 2016-08-26
ft_sql_transformer not working as described


AlexIoannides:
Hi,

`ft_sql_transformer` has the following prototype:

``` r
ft_sql_transformer <- function(x,
                               input_col = NULL,
                               output_col = NULL,
                               sql)
```

But it **doesn't** make use of `input_cols` and `output_cols` in the body. Is this in anticipation of forthcoming functionality or just an oversight?

Either way, I'd be happy to make the changes if you let me know.

`ft_sql_transformer` + HiveQL is an awesome alternative to `dplyr::mutate`. Need to get it out there with a few examples too (again, happy to do it).

kevinushey:
Thanks; this is indeed an oversight -- we should only accept `x` and `sql` for this transformer.

AlexIoannides:
After taking a look at this I **do not** think this is an oversight. 

If `input_cols` and `output_cols` are removed then the following code doesn't work,

```r
iris <- copy_to(sc, iris)
iris %>%
  sdf_mutate(Species = ft_sql_transformer(Species, sql = "SELECT upper(Species) FROM __THIS__"))
```

The variables are needed for `ft_sql_transformer` to work within `sdf_mutate`.

jmuhlenkamp:
+1

PR #943, opened 2017-08-16
Remove unused arguments from ft_sql_transformer

javierluraschi:
See https://github.com/rstudio/sparklyr/issues/194, is this reasonable @kevinushey?


kevinushey:
LGTM! The failing test can be tweaked to ignore the interface change for this particular function.

PR #969, opened 2017-08-25
Allow additional livy session parameters

andresperezl:
When creating a Livy session, it allows passing additional parameters, such as setting the proxyUser, numCores, etc. https://github.com/apache/incubator-livy/blob/master/docs/rest-api.md#post-sessions

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,6 +55,9 @@
   components to extract. Also implemented `sdf_project()` to project datasets using
   the results of `ml_pca()` models.
 
+- Added support for additional livy session creation parameters using
+  the `livy_config()` function.
+
 # Sparklyr 0.6.1 (UNRELEASED)
 
 - Fixed error in `spark_apply()` that may triggered when multiple CPUs
--- a/R/connection_spark.R
+++ b/R/connection_spark.R
@@ -143,8 +143,7 @@ spark_connect <- function(master = "local",
                             app_name,
                             version,
                             hadoop_version ,
-                            extensions,
-                            ...)
+                            extensions)
   } else if (method == "gateway") {
     scon <- gateway_connection(master = master, config = config)
   } else if (method == "databricks") {
--- a/R/connection_spark.R
+++ b/R/connection_spark.R
@@ -143,7 +143,8 @@ spark_connect <- function(master = "local",
                             app_name,
                             version,
                             hadoop_version ,
-                            extensions)
+                            extensions,
+                            ...)
   } else if (method == "gateway") {
     scon <- gateway_connection(master = master, config = config)
   } else if (method == "databricks") {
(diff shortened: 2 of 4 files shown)


javierluraschi:
@andresperezl thanks for this PR, various community members have requested a subset of the parameters you are providing so this is definetely helping the community.

Regarding the proposed changes, we do want to avoid changing `spark_connect()` and instead, we want to set the parameters you are providing under the `config` object. The `spark_connect()` accepts only common parameters across all connections, while the `config` object was designed to be overloaded with many more. **Could you move these parameters into the config object?**

In retrospect, I'm almost sure we are not using the preffix `livy.` see [this](https://github.com/rstudio/sparklyr/blob/master/R/livy_connection.R#L126). So my recommendation would be to remove [this](https://github.com/rstudio/sparklyr/blob/master/R/livy_connection.R#L126) line and reuse this preffix for the Livy session configuration settings you are adding.

Users would set the new parameters as:

```r
config$proxyUser = "foo"
config$jars = "/jar1"
config$driverMemory = "16g"
sc <- spark_connect(master = "<livy>", method = "Livy", config = config)
```

If you want to go the extra mile, you could also provide parameters under `livy_config`, to allow users to set them as:

```
config <- livy_config(
  proxy_user = "foo",
  jars = "/jar1",
  driver_memory = "16g"
)
```

Notice that we use `snake_case` for all our parameter names.

Again, @andresperezl, thank you for this contribution. Happy to merge this one once this feedback is addressed or help out with these changes if you are running low on time. Let us know!

andresperezl:
@javierluraschi thanks for the feedback, I will make the changes. 
Just one note and is about the snake_case, and is that the parameters that Livy accepts in the request POST body are camelCase. See https://github.com/apache/incubator-livy/blob/master/docs/rest-api.md#post-sessions

andresperezl:
I have done what you suggested and used the config object to store the parameters, as well as modifying the livy_config function, so the user can pass the values, and it will convert it automatically to camelCase which is what Livy accepts

javierluraschi:
@andresperezl I would remove the dependency to the `snakecase` package and rather add a simple mapping like the following, we also need an entry to the `NEWS` file. The reasoning is to minimize dependencies since any package that breaks could also break `sparklyr`. Other than that, looks good to me to merge. Thank you again.

```r
params_map <- c(
  proxy_user = "proxyUser,
  jars = "jars,
  py_files = "pyFiles",
  files = files,
  driver_memory = "driverMemory",
  driver_cores = "driverCores,
  executor_memory = "executorMemory",
  executor_cores = "executorCores",
  num_executors = "numExecutors",
  archives = "archives",
  queue = "queue",
  name = "name",
  heartbeat_timeout = "heartbeatTimeout"
)
```

andresperezl:
@javierluraschi I remove the snakecase package and did a mapping as you suggested. Also, added to the NEWS file the support for livy sessions parameters. Let me know if there is anything else I missed. And thanks again for the feedback

javierluraschi:
Thank you, @andresperezl

Issue #1035, opened 2017-09-25
`spark_apply()` hangs for 16min when multiple executors run on same node with package distribution


javierluraschi:
When packages are distributed, fix from `sparkly 0.6.3` to avoid deadlocks while extracting packages is causing a `16min` delay when the lock does trigger caused by incorrect wait time for the lock file to be removed.

See for instsance:

<img width="665" alt="screen shot 2017-09-25 at 1 37 38 pm" src="https://user-images.githubusercontent.com/3478847/30829896-0a7d6dba-a1f7-11e7-9763-d74fb9a0fc72.png">

<img width="667" alt="screen shot 2017-09-25 at 1 37 46 pm" src="https://user-images.githubusercontent.com/3478847/30829895-0a765278-a1f7-11e7-8adc-70d7f44deaea.png">

javierluraschi:
Resolved with https://github.com/rstudio/sparklyr/pull/1036

PR #1036, opened 2017-09-25
`spark_apply()` hangs for 16min when multiple executors run on same node with package distribution

javierluraschi:
See https://github.com/rstudio/sparklyr/issues/1035 and https://stackoverflow.com/posts/46410425/edit - Patch candidate for CRAN.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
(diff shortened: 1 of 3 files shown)


javierluraschi:
Ran #1035 over `10 x m3.xlarge` instances in EMR, looking much better (first run was computing only `nrow()` followed by `myFunction()`):

screen shot 2017-09-25 at 1 59 57 pm

PR #1044, opened 2017-09-28
Note expected input type

russellpierce:
The documentation for input type doesn't specify whether it is NSE or SE.  SE is probably the expected default, but because this package is from Rstudio sometimes one expects NSE.  Regardless, this arg could have been numeric positional or any other number of things.  So, explicit being better than implicit... I've made a minor tweak to the documentation to specify that the value for `partition_by` is expected to be a character vector.


javierluraschi:
@russellpierce makes sense, thank you.

PR #1048, opened 2017-09-29
Update sample_n error comment for Spark < 2.x

FavioVazquez:
I noticed that with Spark < 2.x the function sdf_sample will work. So this could be a more useful suggestion for the user.


FavioVazquez:
Blame @jared275.

javierluraschi:
@FavioVazquez thank you. In the future, you can also regenerate the docs by installing roxygen2 and re-documenting the project, I'll rebuild the docs myself in master.

PR #1107, opened 2017-11-03
Add support in spark_apply() for context param

javierluraschi:
Currently, all closures in `spark_apply()` need to be self-contained and therefore, the following operation is not supported:

```r
a <- 1
spark_apply(sdf_len(sc, 1), function(e) {
    e + a
})
```

To mitigate this and support similar use cases, this PR adds a new `context` parameter to pass an additional object, as in:

```r
spark_apply(sdf_len(sc, 1), function(e, c) {
    e + c
}, context = 1)
```
```
# Source:   table<sparklyr_tmp_174e249c63036> [?? x 1]
# Database: spark_connection
     id
  <dbl>
1     2
```

Or even a function, etc...

```r
spark_apply(sdf_len(sc, 1), function(e, c) {
    e + c(1, 1)
}, context = function(a, b) { a + b })
```
```
# Source:   table<sparklyr_tmp_174e21157de2e> [?? x 1]
# Database: spark_connection
     id
  <dbl>
1     3
```


kevinykuo:
@javierluraschi are we planning to change the API once the `rlanglabs` work is complete? The current implementation bothers me a little because `base::c()`. It'd be nice to have something like
```
spark_apply(sdf_len(sc, 1), function(e, c) {
    e + a
}, context = {a <- 1})
```
But if it's gonna change next month anyway we're good for now.

wkdavis:
@kevinykuo @javierluraschi The documentation for the `spark_apply()` function doesn't provide any background on the `context` argument, and the param definition "Optional object..." isn't clear that any object(s) passed must be wrapped in `{}`. Would you consider adding the examples above to the official documentation to provide more clarity on how the `context` argument is to be implemented?

Issue #1101, opened 2017-10-31
spark_write_source not appropriate for generic Spark data sources


martinstuder:
https://github.com/rstudio/sparklyr/blob/491194f64e92bc5985961e99f65a04fa9d5dffb6/R/data_interface.R#L429

For a generic data source `spark_write_source` (`
spark_data_write_generic`) should call `.format(source)` where `source` is the name of a class extending `org.apache.spark.sql.sources.CreatableRelationProvider` similar to `spark_read_source` where `source`  is the name of a class extending `org.apache.spark.sql.sources.RelationProvider`.

With the current implementation, the source class name is used as method name and leads to a `java.lang.IllegalArgumentException` being thrown.

PR #1109, opened 2017-11-07
fixes rstudio/sparklyr#1101

martinstuder:
Fixes the signature of `spark_write_source` and updates `spark_data_write_generic` to invoke a `save` when using the `format` method on a Spark DataFrame.

### DIFF
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -430,7 +430,9 @@ spark_data_write_generic <- function(df,
     invoke(options, fileMethod, url, path, properties)
   }
   else {
-    invoke(options, fileMethod, path)
+    options <- invoke(options, fileMethod, path)
+    # Need to call save explicitly in case of generic 'format'
+    if(fileMethod == "format") invoke(options, "save")
   }
 
   invisible(TRUE)
@@ -693,7 +695,6 @@ spark_write_jdbc.spark_jobj <- function(x,
 #'
 #' @export
 spark_write_source <- function(x,
-                               name,
                                source,
                                mode = NULL,
                                options = list(),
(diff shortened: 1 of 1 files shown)


R/data_interface.R
```diff
@@ -430,7 +430,9 @@ spark_data_write_generic <- function(df,
     invoke(options, fileMethod, url, path, properties)
   }
   else {
-    invoke(options, fileMethod, path)
+    options <- invoke(options, fileMethod, path)
+    # Need to call save explicitly in case of generic 'format'
+    if(fileMethod == "format") invoke(options, "save")
```
javierluraschi: Thanks much for pushing a fix here!

Should we allow users to also specify the save location? For instance, looks like `avro` follows this path:

https://github.com/databricks/spark-avro#scala-api

```
df.write.format("com.databricks.spark.avro").save("/tmp/output")
```

So we could change your statement to:

```
if(fileMethod == "format") {
   if (is.null(path)) invoke(options, "save") else invoke(options, "save", path)
}
```

which would also mean we keep the `name` parameter in `spark_write_source` and pass a new parameter to specify that this is a format-based write, maybe with a new `is_format = TRUE` parameter.

If you are short on time, I'll merge this fix and push an update to include the path, but either way, would be good to hear your thoughts on the usefulness of including the `path`.

Which data source are you working against? I would think `path` is required in most of them, but I guess if its Cassandra the destination table is specified under `options`?

Thanks!

martinstuder: Hi Javier, thank you for your feedback! 

We're working on Google BigQuery support as a sparklyr extension package (the GitHub repo is not public yet but hopefully soon). As such I was only looking into `org.apache.spark.sql.sources.CreatableRelationProvider` and related interfaces where you supply the destination (table) through `options` as you mentioned.

Supporting `path` for other formats (e.g. file-based ones such as avro) definitely makes sense. I don't know how to support that best with the current `spark_write_source` and `spark_data_write_generic` interfaces without looking into it in more detail.

Having a quick look it seems to me that `name` in `spark_write_source` in that case should rather be called `path` (although that would also imply an interface change). Further, `spark_data_write_generic` either gets an additional parameter `source` to specify the data source class or package name, or an additional flag `is_format` to specify whether this is a format-based write or not. In the latter case, I guess, you would have to (mis-)use `fileMethod` to specify the data source class or package name which might be confusing.

If you like I can have a stab at it tomorrow and send an updated pull request. In that case, let me know if you prefer to keep the existing `spark_write_source` interface unchanged and whether you have a preference for the `spark_data_write_generic` interface.

javierluraschi: Since you are confirming you don't need this parameter for BigQuery, we can merge as it is and consider further improvements later. Thank you, @martinstuder!

martinstuder: Just for reference, here is the link to the package I mentioned https://github.com/miraisolutions/sparkbq

PR #1167, opened 2017-12-11
enable inputting read schema to optimize reads of nested parquet data

mattpollock:
Pending https://github.com/apache/spark/pull/16578 spark reads too much data from nested parquet fields. This PR enables users to pass in read schemas (a sub-schema from the full data schema) so that only the nested (and top level) fields of interest are read in. For users with complex nested data this amounts to a huge performance boost.

The `schema` argument can be constructed using the `sparklyr.nested` functions defined here https://mitre.github.io/sparklyr.nested/reference/struct_type.html (which mostly just use the `SQLUtils.getSQLDataType`  from https://github.com/rstudio/sparklyr/blob/master/inst/livy/spark-1.5.2/sqlutils.scala). 

For example, referencing the schema here https://mitre.github.io/sparklyr.nested/index.html then one could input
```
start_point_schema <- struct_type(sc,
    struct_field(sc, 'altitude', double_type(sc), nullable=FALSE)
)
phases_schema <- struct_type(sc,
  struct_field(sc, 'start_point', start_point_schema, FALSE)
)
data_schema <- struct_type(sc,
  c(struct_field(sc, 'primary_key', string_type(sc), FALSE),
    struct_field(sc, 'phase_sequence', string_type(sc), TRUE),
    struct_field(sc, 'phases', phases_schema, TRUE))
)

df <- spark_read_parquet(sc, ..., schema=data_schema)
```

This would result in not reading any of the `phasts$end_point` data and would read only `phases$start_point$altitude`. With most of the data filtered at read, then doing explode operations (e.g., https://mitre.github.io/sparklyr.nested/reference/sdf_explode.html) is not nearly so expensive.

### DIFF
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -374,7 +374,7 @@ spark_data_read_generic <- function(sc, source, fileMethod, readOptions = list()
     columnDefs <- spark_data_build_types(sc, columns)
   }
   if (readSchemaProvided || columnsHaveTypes) {
-    options <<- invoke(options, "schema", columnDefs)
+    options <- invoke(options, "schema", columnDefs)
   }
 
   df <- invoke(options, fileMethod, source)
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -197,6 +197,7 @@ spark_write_csv.spark_jobj <- function(x,
 #'
 #' @inheritParams spark_read_csv
 #' @param options A list of strings with additional options. See \url{http://spark.apache.org/docs/latest/sql-programming-guide.html#configuration}.
+#' @param schema A (java) read schema. Useful for optimizing read operation on nested data.
 #'
 #' @details You can read data from HDFS (\code{hdfs://}), S3 (\code{s3n://}), as well as
 #'   the local file system (\code{file://}).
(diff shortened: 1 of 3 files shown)


mattpollock:
As far as I can tell this is barking because I didn't add any tests. I Believe adding a test would entail adding a small nested parquet file to the test/resources dir. Do you want me to do that with something small like `iris`? I noted that there is not currently (that I can find) a test for `spark_read_parquet`

javierluraschi:
@mattpollock here are the specific failures, you can find them at the very bottom of Travis, after expanind `552.76s$ tail -n 500 /tmp/sparklyr.log`: 

```
1. Error: spark_read_json() can load data using column types (@test-read-write.R#44) 
cannot change value of locked binding for 'options'
1: spark_read_json(sc, name = "iris_json_typed", path = "test.json", columns = list(Sepal_Length = "character", 
       Species = "character", Other = "struct<a:integer,b:character>")) %>% collect() at testthat/test-read-write.R:44
2: eval(lhs, parent, parent)
3: eval(lhs, parent, parent)
4: spark_read_json(sc, name = "iris_json_typed", path = "test.json", columns = list(Sepal_Length = "character", 
       Species = "character", Other = "struct<a:integer,b:character>"))
5: spark_data_read_generic(sc, spark_normalize_path(path), "json", options, columns)
2. Error: spark_read_text() and spark_write_text() read and write basic files (@test-read-write.R#79) 
cannot change value of locked binding for 'options'
1: spark_read_text(sc, name = "test_spark_read", path = "test.txt") at testthat/test-read-write.R:79
2: spark_data_read_generic(sc, spark_normalize_path(path), "text", options, columns)
```

I think one of the problems is this line:

```
if (readSchemaProvided || columnsHaveTypes) {
 +    options <<- invoke(options, "schema", columnDefs)
    }
```

Since the `<<-` operator is being used, the `base::options` function is trying to be modified, which fails. Changing this to `<-` should fix both failures.

mattpollock:
Thanks @javierluraschi. Fixed now.

mattpollock:
[read_schemas_example.zip](https://github.com/rstudio/sparklyr/files/1563294/read_schemas_example.zip)
See attached for a working example (using `sparklyr.nested`) of what this PR enables.

javierluraschi:
Thank's @mattpollock, merged!

PR #1174, opened 2017-12-19
Update documentation on s3 integration

nhick:
@javierluraschi kindly review.
Updated docs to reflect latest recommendations as per [hadoop  wiki](https://wiki.apache.org/hadoop/AmazonS3) and added some pointers on how to configure spark to keep spark and hadoop config separate from sparklyR.


javierluraschi:
Thank you, @nhick!

PR #1203, opened 2018-01-12
replace jsonlite base64_enc by base64encode from package base64enc

sipemu:
The problem description can be found in the following issue of the jsonlite package:
https://github.com/jeroen/jsonlite/issues/220

Short: The jsonlite encoder adds \n when the string "username:password" is long.


javierluraschi:
Thank you, @sipemu!

PR #1216, opened 2018-01-16
Add support to override compilation of scala object with higher version

javierluraschi:
I noticed that it's currently not possible to override classes during compilation... For instance, one might want to have the following package:

spark-1.5.0/main.scala
spark-1.5.0/foo.scala
spark-2.0.0/foo.scala

This PR adds support to override sources in such a way which could be useful in cases where the Spark API has newer objects, say while using `mllib.*` deprecated over `ml.*`. In this case, 1.5.0 will compile main and foo from the same source dir, but on 2.0.0 sparklyr would compile main form 1.5.0 and foo form 2.0.0 to enable such overrides.


kevinykuo:
I think we may need to rename `mlutils.scala` in the 2.0 folder since there's another file with the same name in the 1.6 folder whose functions are required for both 1.6 and 2.x.

Issue #1127, opened 2017-11-16
Investigate test failures in 1.6.x


kevinykuo:
@javierluraschi could you take a look at these? They came up locally running Spark 1.6.0.

1. test-read-write.R:79: error: `spark_read_text() and spark_write_text() read and write basic files
java.lang.Exception: No matched method found for class org.apache.spark.sql.DataFrameReader.text`
2. test-read-write.R:120: error: `spark_write_table() can append data
org.apache.spark.sql.execution.QueryExecutionException: FAILED: Execution java.net.URISyntaxException: Relative path in absolute URI: file:./writetableappend)`
3. test-serialization.R:163: error: `collect() can retrieve all data types correctly
org.apache.spark.sql.AnalysisException: cannot recognize input near 'integer' ')' 'as' in primitive type specification; line 1 pos 89`
4. test-serialization.R:210: error: `collect() can retrieve NULL data types as NAs
org.apache.spark.sql.AnalysisException: cannot recognize input near 'integer' ')' 'as' in primitive type specification; line 1 pos 98`
5. test-spark-apply.R:130: error: `'spark_apply' works over empty partitions
java.lang.IllegalArgumentException: requirement failed`

javierluraschi:
Worth considering creating 1.6 / 2.0 Travis test configs.

javierluraschi:
A few more...

1. Error: ml_model_kmeans can be used with ml_predict() (@test-ml-clustering-
java.lang.IllegalArgumentException: Could not parse formula: ~. - Species
2. Failure: ft_chisq_select() works properly (@test-ml-feature-chisq-selector
dplyr::pull(result, selectedFeatures) not equal to list(18, 12, 15).
3. Error: ft_max_abs_scaler() works properly (@test-ml-feature-max-abs-scaler
java.lang.ClassNotFoundException: org.apache.spark.ml.feature.MaxAbsScaler
4. Error: 'spark_apply' can filter data.frame (@test-spark-apply.R#130)  ────
java.lang.IllegalArgumentException: requirement failed
```
	at scala.Predef$.require(Predef.scala:221)
	at org.apache.spark.sql.catalyst.analysis.UnresolvedStar.expand(unresolved.scala:199)
	at org.apache.spark.sql.catalyst.analysis.Analyzer$ResolveReferences$$anonfun$apply$10$$anonfun$applyOrElse$14.apply(Analyzer.scala:353)
[...]

PR #1226, opened 2018-01-17
Add support for Spark 1.6.X on Travis

javierluraschi:
See https://github.com/rstudio/sparklyr/issues/1127


kevinykuo:
@javierluraschi let's rebase against master and see how we do

kevinykuo:
Oh just kidding, was looking at the wrong build and saw some old CRAN warnings

PR #1236, opened 2018-01-17
Fix a typo and add missing docs

kevinykuo:
Also attempt to set `warnings_are_errors: true` in travis since builds have been passing with WARNINGs.


R/dbi_spark_table.R
```diff
@@ -12,7 +12,7 @@ setMethod("dbWriteTable", "spark_connection",
 
     if (identical(temporary, FALSE))
       spark_write_table(
-        tbl(sc, temp_name),
+        tbl(conn, temp_name),
```
javierluraschi: Thanks!

Issue #1329, opened 2018-03-01
current design of sdf_mutate makes extensions awkward


mattpollock:
Currently `sdf_mutate` constructs a call such that
```r
df %>%
  sdf_mutate(out=some_function(in1, in2))
```
actually calls `some_function(in1, out, in2)` internally with unnamed arguments in the first two positions (`in2` in this case would keep its name if it had one). This works OK for `ft_*` functions that take a single column input. However it makes it difficult to design functions intended to be used from `sdf_mutate` that have multiple inputs and may refer to multiple columns of the input data frame.

For example, take the following `ft_starts_with` function that can call `startsWith(Column other)` or `startsWith(String literal)` from the [Spark Column API](https://spark.apache.org/docs/2.1.0/api/java/org/apache/spark/sql/Column.html#startsWith(org.apache.spark.sql.Column))
```r
ft_starts_with_v1 <- function(.data, col, output_col, pattern, ...) {
  pattern <- deparse(substitute(pattern))
  ft_starts_with_(.data, col=col, pattern=pattern, output_col=output_col, ...)
}


ft_starts_with_v2 <- function(.data, col, pattern, ..., output_col) {
  
  ft_starts_with_(.data, col=col, pattern=pattern, output_col=output_col, ...)
}

ft_starts_with_ <- function(.data, col, pattern, ..., output_col) {
  
  # get java refs
  sc <- spark_connection(.data)
  sdf <- spark_dataframe(.data)
  
  # function specific error checking
  if (!is.character(col) || !(col %in% colnames(.data)))
    stop("'col' must be a column of the input dataframe")
  col <- invoke(sdf, "col", col)
  
  if (!is.character(pattern))
    stop("'pattern' must either be a character string or else the name of a column")
  if (pattern %in% colnames(.data))
    pattern <- invoke(sdf, "col", pattern)
  
  # call udf - here I'm using the Spark Column API, it might also be a custom scala method
  spk_output_col <- invoke(col, "startsWith", pattern)
  df <- invoke(sdf, "withColumn", output_col, spk_output_col)
  sdf_register(df)
}
```

The current implementation leads to some unexpected results
```r
df <- copy_to(sc, data.frame(a=c("abc", "def", "ghi"), b=c("ab", "ab", "gh")))

df %>% sdf_mutate(sw=ft_starts_with_v1(a, pattern=b)) # works
df %>% sdf_mutate(sw=ft_starts_with_v1(a, pattern="ab")) # works
[...]

Issue #1267, opened 2018-01-24
Consider NSE/tidyselect for exported functions


kevinykuo:
E.g.
```
iris_tbl %>%
  ft_string_indexer(Species, label) %>%
  ft_vector_assembler(vars(starts_with("Petal")), features)
```
cc @edgararuiz

PR #1330, opened 2018-03-01
make sdf_mutate friendlier and more predictable

mattpollock:
This resolves issue #1329


mattpollock:
The documented example works without any chnges
```r
data(beavers, package = "datasets")
beaver_tbl <- copy_to(sc, beaver1, "beaver")
beaver_tbl %>%
  mutate(squared = temp ^ 2) %>%
  sdf_mutate(warm = ft_binarizer(squared, 1000)) %>%
  sdf_register("mutated")
#
#  # Source:   table<mutated> [?? x 6]
# # Database: spark_connection
#      day  time  temp activ squared  warm
#    <dbl> <dbl> <dbl> <dbl>   <dbl> <dbl>
#  1   346   840  36.3     0    1320  1.00
#  2   346   850  36.3     0    1321  1.00
#  3   346   900  36.4     0    1321  1.00
#  4   346   910  36.4     0    1326  1.00
#  5   346   920  36.6     0    1336  1.00
#  6   346   930  36.7     0    1346  1.00
#  7   346   940  36.7     0    1348  1.00
#  8   346   950  36.8     0    1351  1.00
#  9   346  1000  36.8     0    1355  1.00
# 10   346  1010  36.9     0    1360  1.00
# # ... with more rows
```

R/sdf_interface.R
```diff
@@ -271,16 +271,13 @@ sdf_mutate_ <- function(.data, ..., .dots) {
...
       lazy_expr[[1]],        # function
-      data,                  # data
+      data,                  # data,
       output_col=output_col  # output column
-    )
+    ), inputs)               # inputs
 
-    call <- as.call(c(
-      preamble,
-      unlist(inputs)         # parsed (and evaluated if applicable), inputs
```
mattpollock: In the `beavers` example 1000 was getting coerced to "1000" via `unlist` since the other list item was a column name string.

mattpollock:
> The job exceeded the maximum time limit for jobs, and has been terminated.

Not sure what to do about that.

kevinykuo:
@mattpollock Thanks for the PR! Looking at the issue you opened I'm thinking we can consider deprecating `sdf_mutate()` in favor of something like https://github.com/rstudio/sparklyr/issues/1267. Is your use case for `sdf_mutate()` basically to avoid typing a bunch of `"`s? I think we want to avoid having too many ways to achieve the same thing.

mattpollock:
I see your point. I like using the mutate construct where I can name the output column on the LHS and forego the quotes. That makes spark API operations feel more like they belong in a dplyr pipeline. 

I feel pretty strongly that `sdf_mutate` should not stay as is given the odd behavior I pointed out in #1329, however if the direction you all are going in is to promote direct use `ft_*` and similar functions in pipelines I can deal with that. Would #1267 be a breaking change requiring all the current SE functions to be transitioned to NSE and (presumably) new `[ft|ml]_*._$` functions to be written to support standard evaluation?

mattpollock:
I've been mulling this for a 1/2 hour now. I don't presume to have much of a vote here, but keeping NSE in verb functions seems more consistent. Let `ft_` and `ml_` use standard evaluation (i.e., don't do #1267) in the same way that `grepl`, `gsub`, etc. use standard evaluation in local `dplyr` pipelines. Locally most functions use SE and `dplyr` verbs do NSE. Following this pattern seems less confusing with the one caveat being that the `output_col=` argument is unique in the spark context. 

I'd _like_ to keep `sdf_mutate` around with the purposed modifications, but as I said above, if you want to kill it I can modify how I'm doing stuff today.

kevinykuo:
Just gave this some thought, too. The plan for #1267 was that for `ft_.tbl_spark()` functions we'd support looking up unquoted names in the dataframe columns, then look for bindings in the calling environment, so it wouldn't break current usage. However, one of the workflows is that the user would interactively use these `ft_` and `ml_` functions with a data frame, e.g.
```
beaver_tbl %>%
  ft_binarizer("temp", "hot", 36.5) %>%
  ft_r_formula(hot ~ time + activ) %>%
  ml_logistic_regression()
```
then easily turn it into a pipeline by replacing the data frame input:
```
pipeline <- ml_pipeline(sc) %>%
  ft_binarizer("temp", "hot", 36.5) %>%
  ft_r_formula(hot ~ time + activ) %>%
  ml_logistic_regression()
```

NSE would complicate things here since it won't be obvious where we're looking for the variables first. This would be an argument for not implementing #1267 even if it's backwards compatible. (Also, it adds yet another way to do something...)

Perhaps we should patch `sdf_mutate()` now so it supports named arguments and you're unblocked. Also we could designate `sdf_mutate()` as the one place where NSE is supported for feature transformers. We want to move away from lazyeval though and use tidyeval instead.

@kevinushey @edgararuiz @javierluraschi thoughts here?

kevinykuo:
OK we caught up a bit and this is what we'll do. We'll soft deprecate `sdf_mutate()`, so instead of

```
beaver_tbl %>%
  mutate(squared = temp ^ 2) %>%
  sdf_mutate(warm = ft_binarizer(squared, 1000))
```

we'd write

```
beaver_tbl %>%
  mutate(squared = temp ^ 2) %>%
  ft_binarizer("squared", "warm", 1000)
```

We can still use feature transformers without breaking pipelines. The `ft_` functions aren't SQL so this syntax makes that a bit more explicit. Also, the `sdf_` functions should really map 1:1 to Spark DataFrame functions. We can revisit implementing something in the future but it'd be a different name (or package.)

As for https://github.com/rstudio/sparklyr/issues/1267 we'll put it on hold for now. It'd be pretty involved and if we don't do it right it has the potential to cause much confusion and unexpected behavior.

Thanks again for the issue and your thoughts, @mattpollock !

mattpollock:
Works for me. I'll adjust accordingly. Thanks very much.

PR #1385, opened 2018-04-03
Enable Kerberos authentication with Livy

yutannihilation:
While both Livy and Kerberos are individually supported, using them together is not possible yet.

Currently, `livy_config` considers only Basic authentication. But, if the Livy server is using Kerberos authentication, we need to use "Negotiate (SPNEGO) authentication". If we use curl CLI, `--negotiate` will allow us to use Negotiate. With R/httr, it is available via `httr::authenticate("", "", type = "gssnegotiate")`.

c.f.:
- https://stackoverflow.com/questions/26255964/r-negotiate-authentication-with-rcurl-or-httr
- https://stackoverflow.com/a/38664954

Since it is not as simple as adding one header, it seems better to store it as `sparklyr.livy.auth` config and use it when making a request.

Note that users need to get and cache the ticket priorly, which is described on https://spark.rstudio.com/guides/connections/#kerberos

### DIFF
--- a/R/livy_connection.R
+++ b/R/livy_connection.R
@@ -36,6 +36,7 @@ livy_validate_http_response <- function(message, req) {
 #' @param config Optional base configuration
 #' @param username The username to use in the Authorization header
 #' @param password The password to use in the Authorization header
+#' @param negotiate Whether to use gssnegotiate method or not
 #' @param custom_headers List of custom headers to append to http requests. Defaults to \code{list("X-Requested-By" = "sparklyr")}.
 #' @param ... additional Livy session parameters
 #'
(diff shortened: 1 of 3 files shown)


javierluraschi:
@yutannihilation looks good, see this in one test:

```
Codoc mismatches from documentation object 'livy_config':
livy_config
  Code: function(config = spark_config(), username = NULL, password =
                 NULL, negotiate = FALSE, custom_headers =
                 list("X-Requested-By" = "sparklyr"), ...)
  Docs: function(config = spark_config(), username = NULL, password =
                 NULL, custom_headers = list("X-Requested-By" =
                 "sparklyr"), ...)
  Argument names in code not in docs:
    negotiate
  Mismatches in argument names:
    Position: 4 Code: negotiate Docs: custom_headers
    Position: 5 Code: custom_headers Docs: ...
```

Could you run from RStudio: "Build" then "Document" to rebuild the docs and add the doc changes to your commit?

yutannihilation:
Thanks for pointing out! Hope this works.

javierluraschi:
Thank you, @yutannihilation!

yutannihilation:
Thanks!

PR #1429, opened 2018-04-21
Including an example for ml_kmeans()

samuelmacedo83:
I added a simple example using ml_kmeans. 
If you liked, I can add more examples for the others ml functions


kevinykuo:
Thanks! We need to wrap example code in `\dontrun{}` otherwise they'll fail on CRAN where there's no Spark.

Also, could you use `iris_tbl` to be consistent with the style used in the documentation?

Contributions are always welcome!

samuelmacedo83:
I got it, I'll correct this in the next commit.
Thank you for let me contribute with the code

Issue #1395, opened 2018-04-06
support for Java `char` type


lawremi:
How does one pass Java `char` values from R? 

Example:
```
> invoke_new(con, "java.lang.Character", "f") 
Error: java.lang.Exception: No matched constructor found for class java.lang.Character
```

Not sure how best to support this. Could have a `java_char()` function that wraps a single-character string in R, marking it for conversion to `char`.

Come to think of it, there should probably also be a `java_short()`, `java_long()`, and `java_float()`.

lawremi:
Any thoughts on this? I'm not aware of any workarounds, so it's a blocker for me. I could put together a PR if the above idea sounds reasonable.

kevinykuo:
Taking a look. There are a couple issues here, one is method matching for one-character-string inputs, and the other is that `char`s get automatically serialized back to R character vectors so you can't pass it on to another Scala method using `invoke*`.

lawremi:
Right. The idea (from rjava) is to have constructors like `java_char()` that just slap an S3 class on these types to get dispatch.

kevinykuo:
@lawremi This should work now:
```
> invoke_new(sc, "java.lang.Character", "f") 
[1] "f"
```
Feel free to give us a shout if you run into further issues!

lawremi:
Awesome, thanks!

lawremi:
Turns out that I need to send a `java.lang.Character` to a method. How can I do that, since they are automatically converted to R character vectors? I guess you could support passing these single char strings to `java.lang.Character` parameters. Please let me know if you want a separate issue.

kevinykuo:
It should just work like it does with the `java.lang.Character` constructor, which takes a `char`. Could you give it a try and let us know the result?

lawremi:
Does not seem to work. I think the problem is here https://github.com/rstudio/sparklyr/blob/871d951cb36feae36e932128aa34535a6cff1fa3/java/spark-1.5.2/invoke.scala#L36

It checks that `parameterType` is of type `char` but it must also allow `java.lang.Character`. Otherwise it falls through and checks for whether the `java.lang.String` (from deserialization) is a `java.lang.Character`, which fails.

kevinykuo:
@lawremi could you provide a sample function where this fails, to make sure I'm interpreting the use case correctly? Thanks!

lawremi:
Just something like
```
public static void method(Character c) { }
```
[...]

PR #1439, opened 2018-04-26
Invoke method dispatch improvements

kevinykuo:
This change

- Allows one-character length-one character vectors to be matched to a `Char` parameter.
- Allows integer arguments to be matched to `Short` parameters.
- Allows numeric arguments to be matched to `Long` parameters, but integers are still supported to maintain backwards compatibility.

https://github.com/rstudio/sparklyr/issues/1395


inst/livy/spark-1.5.2/invoke.scala
```diff
@@ -67,6 +79,25 @@ class Invoke {
...
                 // Convert a Java array to scala Seq
                 args(i) = args(i).asInstanceOf[Array[_]].toSeq
+              } else if (parameterTypes(i) == classOf[Char] &&
+                         args(i) != null && args(i).isInstanceOf[String]) {
+                // Convert String to Char
+                args(i) = new java.lang.Character(args(i).asInstanceOf[String](0))
+              } else if (parameterTypes(i) == classOf[Short] &&
+                         args(i) != null && args(i).isInstanceOf[Integer]) {
+                // Convert Integer to Short
+                args(i) = new java.lang.Short(args(i).asInstanceOf[Integer].toShort)
```
javierluraschi: We need the same boundaries check you added but for shorts:

```
// Try to convert Double to Short
val argDouble = args(i).asInstanceOf[Double]
if (argDouble > Short.MaxValue || argDouble < Short.MinValue) {
  throw new Exception("Unable to cast numeric to short: out of range.")
}
```

PR #1455, opened 2018-04-29
An example of ml_pca()

samuelmacedo83:
It's an example of ml_pca(). 
I didn't include an example for ft_pca() because ml_pca() is a wrapper around ft_pca().


javierluraschi:
Thanks, @samuelmacedo83.

PR #1465, opened 2018-05-01
Keeping old Backend.main() shim for databricks mode

falaki:
This patch address issue 1461. Latest sparklyr code refactoring caused two issues:
1. default value for master was removed
2. main class shim in Backend was removed.

This patch fixes both. For the first one I am copying @kevinykuo's fix here. For the second issue I am adding back the Shim.

For some context, Databricks was supposed to switch to switch to new main class last year. Unfortunately, my patch in our runtime got reverted and I did not realize. I am making sure it happens now, but for backwards compatibility it would be good to keep the shim for a bit longer.

cc @javierluraschi

### DIFF
--- a/R/connection_spark.R
+++ b/R/connection_spark.R
@@ -64,7 +64,7 @@ NULL
 #' spark_disconnect(sc)
 #'
 #' @export
-spark_connect <- function(master = "local",
+spark_connect <- function(master,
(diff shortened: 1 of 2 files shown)


javierluraschi:
Looks good! I can regenerate the jars for you. There is a `configure.R` script which bundles all the versions, livy sources and other related changes.

PR #1479, opened 2018-05-07
An example for ml_decision_tree()

samuelmacedo83:
I added an example using multiclass classification


javierluraschi:
Thanks, @samuelmacedo83!

PR #1483, opened 2018-05-10
An example for ml_random_forest()

samuelmacedo83:


samuelmacedo83:
I guess the problem is in spark version 2.3.0. 
When I run `spark_connect(master = "local")` in past versions, it works.

I installed the version 2.3.0 using `spark_instal()l`

javierluraschi:
Yes, @samuelmacedo83, we have a fix for Windows and 2.3.0 here: https://github.com/rstudio/sparklyr/pull/1486 but change looks good, appveyor tests were just recently added and will be fixed shortly.

PR #1489, opened 2018-05-11
WIP: Use JDK7 in travis to ensure backwards compatibility

javierluraschi:


javierluraschi:
This PR is not needed, validated that Spark only supports 1.8 from Spark 2.2+

PR #1498, opened 2018-05-16
Add support for sql editing from connections pane

javierluraschi:
Makes use of https://github.com/rstudio/rstudio/pull/2791 to support a SQL action in the connections pane to edit SQL from an active connection.

screen shot 2018-05-15 at 8 32 31 pm

screen shot 2018-05-15 at 8 29 14 pm

screen shot 2018-05-15 at 8 29 23 pm


harryprince:
SQL panel seems not the main issue, maybe RStudio fails to run multi-query is more important than that.

Currently, Rmarkdown SQL chunk is good enough.

PR #1518, opened 2018-05-24
adds support for printing the RDD debug string 

mattpollock:
This mimics the pySpark `toDebugString` method. Useful for diagnosing how spark is splitting a pipeline up into stages and also for inspecting the parallelization (number of partitions) at each stage.

### DIFF
--- a/R/spark_utils.R
+++ b/R/spark_utils.R
@@ -56,3 +56,18 @@ jobj_class <- function(jobj, simple_name = TRUE) {
                 ensure_scalar_boolean(simple_name)) %>%
     unlist()
 }
+
+#' Description of dependencies for debugging
+#' 
+#' Print a plan of execution to generate \code{.data}. This plan will, among other things, show the
+#' number of partitions in parenthesis at the far left and indicate stages using indendatations.
+#' 
+#' @param x An \R object wrapping, or containing, a Spark DataFrame.
+#' @export
+spark_debug_string <- function(x) {
+  spark_dataframe(x) %>% 
+    invoke("rdd") %>% 
+    invoke("toDebugString") %>% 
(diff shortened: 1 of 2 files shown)


javierluraschi:
@mattpollock this is a great addition! One suggestion, it currently prints like:

screen shot 2018-05-24 at 10 29 19 am

But it would be nicer to print with `cat` inside `spark_debug_string`, with something that would look like:

screen shot 2018-05-24 at 10 33 28 am

I can make this change if you are running short on time, let me know.

mattpollock:
Yeah, I agree. I've been using `cat()` locally. Do you think there might be a use case where someone wants to capture the output to a character vector and programatically do something like determine the choke point (stage with min number of partitions)? That sort of thing will be more annoying with `cat`.

mattpollock:
Hmmm, this an #1519 are failing in travis for Spark 2.3 on the exact same error 

```
/home/travis/.travis/job_stages: eval: line 57: syntax error near unexpected token `('
/home/travis/.travis/job_stages: eval: line 57: `Rscript -e covr::codecov() '
```

despite my branching off master and making completely independent changes for both PRs. Is something in the CI job busted or am I doing something really stupid here?

javierluraschi:
@mattpollock we've seen some intermittent issues in Travis, I'll just rerun the job to get a clean pass before merging this one.

javierluraschi:
@mattpollock renamed `spark_debug_string()` to `sdf_debug_string()`, all our Spark DataFrame operations are prefixed `sdf_`, see https://github.com/rstudio/sparklyr/commit/e679cb129afe89d8b3b65b281ef8cdb396abc2d1.

mattpollock:
Thanks!

PR #1519, opened 2018-05-24
add dynamic config setter/getter

mattpollock:
I find this useful for functions that define spark jobs but take `sc` as an argument. If there are `spark_read_*` operations in the function and other params (e.g., number of days or directories of data) that can be translated to a partitioning scheme then adjusting that field via `spark_set_config` and then putting it back again at the end of the function enables the function author to optimize the operation without owning the spark connection.


mattpollock:
@javierluraschi do you understand why travis failed in for spark 2.3.0? The syntax error message doesn't mesh with passing for other spark versions and with local testing. I'm happy to fix - I just have no idea what needs fixing at the moment.

javierluraschi:
@mattpollock for consistency with `spark_hive_config()`, `spark_context_config()` I renamed `spark_get_config()` and `spark_set_config()` to `spark_session_config()`, see https://github.com/rstudio/sparklyr/commit/aceb904cf54605c4b97d103b27969630f8eef166.

mattpollock:
Thanks much @javierluraschi

PR #1527, opened 2018-05-29
An example using cross validation

samuelmacedo83:


javierluraschi:
Thank you, @samuelmacedo83, nice work getting so many `sparklyr` examples to the docs.

samuelmacedo83:
Thank you @javierluraschi. It's my pleasure to contribute with the package because I work with it every day and help me a lot.

Issue #1530, opened 2018-06-01
spark_apply set .libPaths


mattpollock:
I know that a lot of work has gone into making packages available to R workers living in spark executors. However one option that does not seem to be supported is simply setting `.libPaths` to something.

Naturally users _could_ do
```
tbl %>%
  spark_apply(function(x, context){.libPaths("something"), ...})
```
but it would be nice to enable a system level configuration (think RStudio Server running on a cluster edge node). Would it be possible to support something similar to `spark.r.command` (e.g., `spark.r.libpaths`) which, if set, will cause `.libPaths(config$spark.r.libpaths)` to be executed before `f` in the workers.

mattpollock:
Maybe using `sparklyr.apply.options.rscript.before` is the right approach here. I can't find documentation on how to use that though. Should the value of this param contain R code to be executed or should it point to a file? If it is a file path then where should the file live (e.g., does `~/before.R` refer to my local home dir or my /user/me dir on HDFS)?

javierluraschi:
@mattpollock that sounds reasonable, here is a PR: https://github.com/rstudio/sparklyr/pull/1533

I haven't tested this at all, would you mind giving it a shot? You can install with:

```r
devtools::install_github("rstudio/sparklyr", ref = "bugfix/spark-apply-rlibpaths")
```

mattpollock:
Thanks @javierluraschi! I'll take a look.

mattpollock:
resolved with #1533

PR #1533, opened 2018-06-04
Support configuring library path in spark_apply()

javierluraschi:
See https://github.com/rstudio/sparklyr/issues/1530


mattpollock:
Just tested this in a YARN environment and it worked great. My only suggestion is that the default value of `packages=` depend on whether `spark.r.libpaths` is set. If it is not, then the current `packages=TRUE` default seems good. If it is however then much of the point of specifying the configuration is to avoid the copy operation. If `packages=TRUE` then there are likely to be frustrated users who forget to set `packages=FALSE` for every call to `spark_apply`

mattpollock:
I would actually prefer `packages=FALSE` as a general default (make users think about it before their entire `.libPaths()` dirs are copied up to each executor), but recognize that this would be a breaking change for some users.

javierluraschi:
@mattpollock good to hear, and yes, good suggestion. I'll implement this with a new configuration parameter `sparklyr.apply.packages` which you can see to `FALSE` but others could set to a subset of packages too copy, etc.

mattpollock:
Looks good to me. Thanks for a quick response on this feature add! I'm happy to kill my clunky work around capturing of `f` and wrapping it in a function that calls `.libPaths`.

Issue #1538, opened 2018-06-07
Remove unnecessary sleep to improve performance


ciasom:
In function livy_post_statement, the while loop has an unnecessary sleep even when statementReponse$state has already changed to "available".

```
  while ((statementReponse$state == "running" || statementReponse$state == "waiting" ) &&
         waitTimeout > 0) {
    statementReponse <- livy_get_statement(sc, statementReponse$id)

    Sys.sleep(sleepTime)

    waitTimeout <- waitTimeout - 1
    sleepTime <- sleepTime * 2
  }
```

An immediate check like the following should be added after  livy_get_statement to avoid the last sleep. For a big query, this can save hours.

```
   if((statementReponse$state != "running" && statementReponse$state != "waiting" ) || waitTimeout <= 1)
     break
```

By adding some print, I got the timing (before change):

livy_post_statement start@2018-06-07 13:35:46
livy_post_statement get_statement (state=waiting) 600 @2018-06-07 13:35:46
livy_post_statement get_statement (state=running) 599 @2018-06-07 13:35:46
livy_post_statement get_statement (state=running) 598 @2018-06-07 13:35:47
livy_post_statement get_statement (state=running) 597 @2018-06-07 13:35:47
livy_post_statement get_statement (state=running) 596 @2018-06-07 13:35:47
livy_post_statement get_statement (state=running) 595 @2018-06-07 13:35:47
livy_post_statement get_statement (state=running) 594 @2018-06-07 13:35:48
livy_post_statement get_statement (state=running) 593 @2018-06-07 13:35:48
livy_post_statement get_statement (state=running) 592 @2018-06-07 13:35:49
livy_post_statement get_statement (state=running) 591 @2018-06-07 13:35:49
livy_post_statement get_statement (state=running) 590 @2018-06-07 13:35:50
livy_post_statement get_statement (state=running) 589 @2018-06-07 13:35:51
livy_post_statement get_statement (state=running) 588 @2018-06-07 13:35:54
livy_post_statement get_statement (state=running) 587 @2018-06-07 13:35:58
livy_post_statement get_statement (state=running) 586 @2018-06-07 13:36:06
livy_post_statement get_statement (state=running) 585 @2018-06-07 13:36:23
livy_post_statement get_statement (state=running) 584 @2018-06-07 13:36:56
livy_post_statement get_statement (state=running) 583 @2018-06-07 13:38:02
livy_post_statement end@2018-06-07 13:40:13

After adding the check, got this:
[...]

PR #1541, opened 2018-06-07
Improve Livy performance for long execution queries

javierluraschi:
Fix for https://github.com/rstudio/sparklyr/issues/1538, fixed `livy.session.command.timeout` and support for  `livy.session.command.interval` to control max polling while waiting for command response.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # Sparklyr 0.9.0 (unreleased)
 
+- Improve performance in Livy for long execution queries, fixed
+  `livy.session.command.timeout` and support for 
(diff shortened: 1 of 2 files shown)


ciasom:
Thanks Javier.  Yes, that fix is what I want. There should be a reasonably small maximum value for the time to sleep.

javierluraschi:
Sounds good.

BTW. My initial fix had a typo, so if you synced to this branch, sync again. I will merge once test succeed.

ciasom:
Hi @javierluraschi, when do you think the next release (including this improvement) will be ready?

PR #1574, opened 2018-06-27
allow passing single-char strings as java.lang.Character args

lawremi:
As requested, this should fix issue (#1395).


javierluraschi:
Thank you, @lawremi!

PR #1583, opened 2018-07-03
An example for ml_gradient_boosted_trees()

samuelmacedo83:
I also change the description because GBTclassifier does not support multiclass yet.


javierluraschi:
@samuelmacedo83 makes sense, thanks!

Issue #1596, opened 2018-07-13
cannot fit 'intercept only' logistic regression model 


shabbybanks:
It is possible to build a 'intercept only' glm in `R`, but not via `ml_logistic_regression`:

```r
# testing
train_data <- iris %>%
  mutate(is_setosa=as.numeric((Species=='setosa'))) %>%
  mutate(is_not_setosa=1-is_setosa) %>%
  setNames(gsub('\\.','_',names(.)))

# runs just fine:
r_mod_one <- glm(formula=is_setosa ~ 1,data=train_data,family=binomial(link='logit'))

# copy data to spark ...
spark_data <- copy_to(sc,train_data,'like_iris',overwrite=TRUE)
# then fit
# but this errors out:
s_mod_one <- spark_data %>% ml_logistic_regression(formula = is_setosa ~ 1)
```

Depending on the size of the data (I was trying this on 'real' data, not `iris`), it may take a while, then throw an error, which suggests the problem is somewhere later in the processing, rather than earlier. The error message is the mystifying:
```
 java.lang.IllegalArgumentException: requirement failed: Vector should
have dimension larger than zero.
```
with this awesome stack trace:
```
 1: spark_data %>% ml_logistic_regression(formula = is_setosa ~ 1)
 2: withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
 3: eval(quote(`_fseq`(`_lhs`)), env, env)
 4: eval(quote(`_fseq`(`_lhs`)), env, env)
 5: `_fseq`(`_lhs`)
 6: freduce(value, `_function_list`)
 7: withVisible(function_list[[k]](value))
 8: function_list[[k]](value)
 9: ml_logistic_regression(., formula = is_setosa ~ 1)
10: ml_logistic_regression.tbl_spark(., formula = is_setosa ~ 1)
11: ml_generate_ml_model(x, predictor = predictor, formula = formula, features_
12: pipeline %>% ml_fit(x)
13: withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
14: eval(quote(`_fseq`(`_lhs`)), env, env)
15: eval(quote(`_fseq`(`_lhs`)), env, env)
16: `_fseq`(`_lhs`)
17: freduce(value, `_function_list`)
18: withVisible(function_list[[k]](value))
19: function_list[[k]](value)
20: ml_fit(., x)
[...]

PR #1597, opened 2018-07-13
get model coefficients, not the function

shabbybanks:
This should fix part of #1596 : fitting a model without an intercept term. The main problem of #1596 is still open, however.

### DIFF
--- a/R/ml_classification_logistic_regression.R
+++ b/R/ml_classification_logistic_regression.R
@@ -369,7 +369,7 @@ new_ml_model_logistic_regression <- function(
         c("(Intercept)", feature_names)
       )
     else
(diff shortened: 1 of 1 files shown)


javierluraschi:
@kevinykuo this one looks like a safe merge since `coefficients` is not defined, but worth double checking in case I missed something.

javierluraschi:
Thank you, @shabbybanks.

kevinykuo:
@shabbybanks Thanks! Could you also contribute a test for this change?

shabbybanks:
ah, ok. I will. Since the code previously threw an error, when now it should not, do you have a preferred `testthat` idiom for testing 'does not throw an error'? Would it be `expect_error(f(), NA)`?

PR #1598, opened 2018-07-13
Bugfix/logistic coef sign

shabbybanks:
sorry for the multiple commits in one PR, feel free to cherry pick the bd1446d commit only.
note that this only adds the test, not the fix! so tests will fail.

### DIFF
--- a/R/ml_classification_logistic_regression.R
+++ b/R/ml_classification_logistic_regression.R
@@ -369,7 +369,7 @@ new_ml_model_logistic_regression <- function(
         c("(Intercept)", feature_names)
       )
     else
-      rlang::set_names(coefficients, feature_names)
+      rlang::set_names(model$coefficients, feature_names)
     coefficients
   }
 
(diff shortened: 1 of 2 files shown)


javierluraschi:
@shabbybanks looks like this change breaks a couple tests:

```
== testthat results  ===========================================================
  OK: 768 SKIPPED: 96 FAILED: 2
  1. Failure: ml_logistic_regression() agrees with stats::glm() for reversed categories (@test-ml-classification-logistic-regression.R#112) 
  2. Failure: ml_logistic_regression() agrees with stats::glm() for reversed categories (@test-ml-classification-logistic-regression.R#120)
```

You can run this tests locally by running:

```
testthat::test_file("tests/testthat/test-ml-classification-logistic-regression.R")
```

javierluraschi:
I see the fix is also included, so the new tests should be passing.

shabbybanks:
I apologize, as I only can use github from the web (corporate something something) and the PR had commits from two different issues. The tests _should_ still be failing because I did not amend the code to deal with the reversed coefficients. I consider this a hidden bug that was not previously being tested. I await your decision on how to approach the issue.

javierluraschi:
@kevinykuo is taking a look at the coefficients reversal issue from https://github.com/rstudio/sparklyr/pull/1598, if a fix is provided, Kevin can consider reusing the tests in the PR. Until then, I'll close this PR and we can follow up on the issue. Regardless, thanks for sending this our way!

PR #1603, opened 2018-07-16
test if we can fit a model without an intercept

shabbybanks:
add test for #1597 : logistic regression without intercept should not error.

### DIFF
--- a/tests/testthat/test-ml-classification-logistic-regression.R
+++ b/tests/testthat/test-ml-classification-logistic-regression.R
@@ -99,6 +99,13 @@ test_that("ml_logistic_regression() agrees with stats::glm()", {
   expect_equal(unname(coef(r)), unname(coef(s)), tolerance = 1e-5)
 })
 
+test_that("ml_logistic_regression can fit without intercept",{
+  expect_error(ml_logistic_regression(
+    iris_weighted_tbl,
+    formula = versicolor ~ Sepal_Width + Petal_Length + Petal_Width,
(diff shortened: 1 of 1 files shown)


kevinykuo:
Is it possible to compare the fitted coefficients against `stats::glm()`?

shabbybanks:
OK, I just added a check.

tests/testthat/test-ml-classification-logistic-regression.R
```diff
@@ -99,6 +99,15 @@ test_that("ml_logistic_regression() agrees with stats::glm()", {
   expect_equal(unname(coef(r)), unname(coef(s)), tolerance = 1e-5)
 })
 
+test_that("ml_logistic_regression can fit without intercept",{
+  expect_error(s <- ml_logistic_regression(
+    iris_weighted_tbl,
+    formula = versicolor ~ Sepal_Width + Petal_Length + Petal_Width,
+    fit_intercept=FALSE),NA)
+  r <- glm(versicolor ~ Sepal_Width + Petal_Length + Petal_Width - 1, family=binomial(logit), data=iris_weighted)
```
javierluraschi: I believe you want to change this to:

```
r <- glm(versicolor ~ Sepal.Width + Petal.Length + Petal.Width - 1, family=binomial(logit), data=iris_weighted)
```

shabbybanks:
yes, of course, those were wrong. fixing.

Issue #1591, opened 2018-07-11
coefficients from logistic regression have arbitrary sign?


shabbybanks:
I have found that _sometimes_ the sign of coefficients from `ml_logistic_regression` are opposite what I would expect, although predictions from the model come out sensibly, which makes me think there is some kind of remapping under the hood. Here is a minimal example illustrating what I would consider a problem:
```r
library(sparklyr)
library(dplyr)
# sc is a spark context.
# make dummy data
train_data <- iris %>%
  mutate(is_setosa=as.numeric((Species=='setosa'))) %>%
  mutate(is_not_setosa=1-is_setosa) %>%
  setNames(gsub('\\.','_',names(.)))
# fit logistic models in R
r_mod_fwd <- glm(formula=is_setosa ~ Sepal_Length,data=train_data,family=binomial(link='logit'))
r_mod_rev <- glm(formula=is_not_setosa ~ Sepal_Length,data=train_data,family=binomial(link='logit'))
# copy data to spark
spark_data <- copy_to(sc,train_data,'like_iris',overwrite=TRUE)
# fit logistic models in spark
s_mod_fwd <- spark_data %>% ml_logistic_regression(formula = is_setosa ~ Sepal_Length)
s_mod_rev <- spark_data %>% ml_logistic_regression(formula = is_not_setosa ~ Sepal_Length)
# get the coefficients:
# in R they are opposites
coefficients(r_mod_fwd)
coefficients(r_mod_rev)
# in spark, not so much:
coefficients(s_mod_fwd)
coefficients(s_mod_rev)
# why?
```
When I run this, I get the output: (sorry for the cut and paste job; `reprex` was being fussy about my spark context)
```
> coefficients(r_mod_fwd)
 (Intercept) Sepal_Length
      27.829       -5.176
> coefficients(r_mod_rev)
 (Intercept) Sepal_Length
     -27.829        5.176
> # in spark, not so much:
> coefficients(s_mod_fwd)
 (Intercept) Sepal_Length
      27.829       -5.176
> coefficients(s_mod_rev)
 (Intercept) Sepal_Length
      27.829       -5.176
```
Logistic regression in `R` does the right thing, flipping the sign of coefficients when I fit the 'opposite' model. However, the fit via `ml_logistic_regression` seems to choose the reference class the same way in both fits, and remaps them later.

I checked, and prediction seems fine here, respecting whatever remapping is going on, so we have
```r
# prediction is fine, of course
[...]

PR #1606, opened 2018-07-17
[WIP] Don't force index numeric labels for classification

kevinykuo:
Fixes https://github.com/rstudio/sparklyr/issues/1591

@shabbybanks I cherry-picked your test from https://github.com/rstudio/sparklyr/pull/1598 and it seems to be passing at least locally. Please give this a shot and propose more tests as needed.

The use case that this change would break is where the labels are two distinct integers that are not 0 and 1, e.g.

```
library(sparklyr)
library(dplyr)
train_data <- iris %>%
  mutate(
    species_type = ifelse(Species == "setosa", 2, 1),
  )
spark_data <- copy_to(sc, train_data, "like_iris", overwrite = TRUE)
s_mod <- spark_data %>%
  ml_logistic_regression(formula = species_type ~ Sepal_Length)
s_mod
# Formula: species_type ~ Sepal_Length
# 
# Coefficients:
#      (Intercept) Sepal_Length
# [1,]   -3.455491 -5.161826758
# [2,]  -12.186625  5.168782843
# [3,]   15.642117 -0.006956085
```

but this is arguably less likely than users being misled by 0 and 1 being mapped to 1 and 0.

Will add a NEWS entry before merge that makes this behavior clear.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
-Version: 0.8.4.9009
+Version: 0.8.4.9010
 Authors@R: c(
     person("Javier", "Luraschi", email = "javier@rstudio.com", role = c("aut", "cre")),
     person("Kevin", "Kuo", role = "aut", email = "kevin.kuo@rstudio.com",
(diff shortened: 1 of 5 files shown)


shabbybanks:
thanks for the fix!

PR #1610, opened 2018-07-18
Example for ml_generalized_linear_regression()

samuelmacedo83:
I used a grid of families and links to compare the rmse


samuelmacedo83:
Hi @javierluraschi . I didn't know if I could do this. I'll do that in the next.
Thank you very much :)

PR #1639, opened 2018-08-15
Improvements when Spark shut downs with OutOfMemory

javierluraschi:
When Spark shut downs with `OutOfMemory` exception, the connection closes and a few side effects trigger in `sparklyr`. The most prominent problem is calling multiple times `cancelAllJobs` which will fail multiple times, additionally, fixing multiple creations of hive context due to delayed creation and missing to register in all connection objects.


R/livy_connection.R
```diff
@@ -787,4 +791,20 @@ initialize_connection.livy_connection <- function(sc) {
   })
 }
 
+#' @export
+hive_context.livy_connection <- function(sc) {
+  if (is.null(sc$hive_context))
+    sc$hive_context <- create_hive_context(sc)
```
kevinykuo: Should this be `sc$state$hive_context`?

javierluraschi: Livy connections are actually eager, so in theory no, see https://github.com/rstudio/sparklyr/blob/69c379d387b7e42e8e38ebb9dfe2dc0df850671e/R/livy_connection.R#L776

kevinykuo: Could we remove these `if` blocks then, since they never do anything?

javierluraschi: In theory, they don't... let's try out and see!

PR #1642, opened 2018-08-16
[TEST] Use kryo serializer by default

javierluraschi:
Running tests against kryo serializer.


javierluraschi:
Looks like this works fine against all our test cases, good to know but not worth the merge, I believe.

Issue #1496, opened 2018-05-15
Hive support breakage in Spark 2.3


HugoHehn:
Following a cluster upgrade I have been experiencing issues with sparklyr not finding Hive databases when connected to the Yarn client. Other than this everything seems to work - it is possible to load local data into a Spark dataframe and run operations on it.

Have done a lot of troubleshooting and searching solutions and most people seem to point to the hive-site.xml file. However, this looks to include all the necessary information and is stored in SPARK_HOME/conf as recommended.

Yarn and Spark seems to be able to find the Hive tables in Shell mode.

Versions used:

- R 3.4.1
- Hive 2.3.2
- Hue 4.1.0
- Spark 2.3.0

Code is basically just the package installation/loading plus:

```

Sys.setenv(SPARK_HOME = '/usr/lib/spark')

sc <- spark_connect(master="yarn-client", config=config, version = "2.3.0", hadoop_version = "2.8.3")

```

Has anyone had this issue?

javierluraschi:
@HugoHehn could you elaborate on what got upgraded? Spark or also sparklyr? Would also help to see the stack trace and logs of the failed operation?

harryprince:
@HugoHehn you can specific the hive-site.xml file in config like this:

```
config$`sparklyr.shell.files` = "/etc/hive/conf/hive-site.xml"
```

it will save your time.

HugoHehn:
@javierluraschi  @harryprince 

Sorry about the late response. Issues still persist.

Cluster was upgraded to a newer EMR version along with newer versions of Spark and Hive. Sparklyr is reinstalled daily so it would also have updated, but it does not seem like that is the issue as reverting to earlier commit does not solve it.

No warning is printed as spark_connect() does not technically fail (it connects to Yarn as it should but does not see any Hive tables) but in the connection log there is a relevant warning:

```
18/05/30 12:18:30 WARN SparkContext: Using an existing SparkContext; some configuration may not take effect.
18/05/30 12:18:31 INFO SharedState: loading hive config file: file:/etc/spark/conf.dist/hive-site.xml
18/05/30 12:18:31 INFO SharedState: Setting hive.metastore.warehouse.dir ('null') to the value of spark.sql.warehouse.dir ('hdfs:///user/spark/warehouse').
18/05/30 12:18:31 INFO SharedState: Warehouse path is 'hdfs:///user/spark/warehouse'.
```

[...]

Issue #415, opened 2017-01-10
Disable hive support


raphael-psr:
Is it possible to disable the hive support in _sparklyr_?

Similar in _sparkR_:
`sparkR.session(master="local[*]", enableHiveSupport=FALSE)`

javierluraschi:
This would make sense for those without Hive enabled. @raphael-psr what is the side effect you are experiencing? I'm asking since `sparklyr` fallsback to no Hive support when it's not available, so in theory, this would work for your use case.

elgalu:
@raphael-psr did you find a way? I'm facing the same issue.

kevinykuo:
The fix for https://github.com/rstudio/sparklyr/issues/1496 should provide a way to do this. Should have a PR within the next day or so.

PR #1646, opened 2018-08-23
Fix Hive support in Spark 2.3

kevinykuo:
- `spark_connect()` now first initializes a `SparkSession` before obtaining a `SparkContext` from it. This is required in order to properly set the static SQL configs such as `spark.sql.catalogImplementation`. The `SparkSession` is immediately cached as the `hive_context` field.
- Default `spark.sql.catalogImplementation` to `hive` so Hive support is enabled. Users can opt out of this by setting the option to `in-memory`.
- `spark_version()` now actually caches the Spark version in the connection object.

Closes https://github.com/rstudio/sparklyr/issues/1496, closes https://github.com/rstudio/sparklyr/issues/415


R/shell_connection.R
```diff
@@ -546,34 +546,82 @@ initialize_connection.spark_shell_connection <- function(sc) {
...
+          }
+        )
+
+        ctx
+      }
 
       invoke(backend, "setSparkContext", sc$spark_context)
     }
 
-    sc$spark_context$connection <- sc
```
javierluraschi: We probably still need this one and the `java_context` bellow.

When we call `invoke_static()` with an `sc`, the `sc` becomes attached to the object, this is to be able to execute `invoke()` on the object etc. This is fine for most operations; however, if you need to get the `spark_context()` and then indirectly get the `spark_context()` from there again, you would get a `spark_context()` that is no longer there. The way we patched this was by reattaching the connection once the `spark_context()` is created.

R/spark_version.R
```diff
@@ -18,8 +18,12 @@ spark_version_clean <- function(version) {
 spark_version <- function(sc) {
 
   # use cached value if available
-  if (!is.null(sc$spark_version))
-    return(sc$spark_version)
+  if (!is.null(sc$state$spark_version))
```
javierluraschi: This is nice! Get's called a bunch of times! I had to do this for Livy already (see https://github.com/rstudio/sparklyr/blob/3a3142d519026e294dc0f50d400cf052c5f8d1a6/R/livy_connection.R#L778),but it doesn't hurt to do this for shell connections as well! 👍

javierluraschi:
Ah, yes, so regarding the broken test, the second code review comment about keeping the re-assignment of `sc` will probably fix this failure:

```
Unable to retrieve a spark_connection from object of class NULL
1: df1_tbl %>% sdf_broadcast() %>% left_join(df2_tbl, by = "b") %>% spark_dataframe() %>% 
       invoke("queryExecution") %>% invoke("optimizedPlan") %>% invoke("toString") at testthat/test-dplyr.R:118
2: withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
3: eval(quote(`_fseq`(`_lhs`)), env, env)
4: eval(quote(`_fseq`(`_lhs`)), env, env)
5: `_fseq`(`_lhs`)
6: freduce(value, `_function_list`)
7: function_list[[i]](value)
8: sdf_broadcast(.)
9: invoke_static(sc, "org.apache.spark.sql.functions", "broadcast", sdf) %>% sdf_register()
10: withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
...
28: `_fseq`(`_lhs`)
29: freduce(value, `_function_list`)
30: function_list[[i]](value)
31: invoke(., "sql", as.character(sqlFields))
32: invoke_trace(spark_connection(jobj), "Invoking", method)
33: spark_config_value(sc$config, "sparklyr.invoke.trace", FALSE)
34: name %in% names(config)
35: spark_connection(jobj)
36: spark_connection.default(jobj)
37: stop("Unable to retrieve a spark_connection from object of class ", paste(class(x), 
       collapse = " "), call. = FALSE)
```

javierluraschi:
The Windows test might be related as well to the previous comment, but not sure, might be something else worth investigating. Here is the output for convenience:

```
-- 1. Error: 'spark_connect' can provide a 'spark_log' (@test-connect-shell.R#15
cannot open the connection
1: capture.output({
       spark_log(sc)
   }) at testthat/test-connect-shell.R:15
2: evalVis(expr)
3: withVisible(eval(expr, pf))
4: eval(expr, pf)
5: eval(expr, pf)
6: spark_log(sc)
7: spark_log.spark_shell_connection(sc)
8: readLines(log)

/usr/bin/tar: Cannot connect to C: resolve failed
/usr/bin/tar: Cannot connect to C: resolve failed
== testthat results  ===========================================================
OK: 1849 SKIPPED: 22 FAILED: 1
1. Error: 'spark_connect' can provide a 'spark_log' (@test-connect-shell.R#15) 
```

R/shell_connection.R
```diff
@@ -546,34 +546,82 @@ initialize_connection.spark_shell_connection <- function(sc) {
...
 
       # create the spark context and assign the connection to it
-      sc$spark_context <- invoke_static(
-        sc,
-        "org.apache.spark.SparkContext",
-        "getOrCreate",
-        conf
-      )
+
+      sc$spark_context <- if (spark_version(sc) >= "2.0") {
```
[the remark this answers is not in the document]
kevinykuo: The "Hive context" for Spark 2.0+ is really the `SparkSession`, which we need anyway, so I don't think we're losing performance here, no?

R/shell_connection.R
```diff
@@ -546,34 +547,82 @@ initialize_connection.spark_shell_connection <- function(sc) {
...
-        sc,
-        "org.apache.spark.SparkContext",
-        "getOrCreate",
-        conf
-      )
 
-      invoke(backend, "setSparkContext", sc$spark_context)
-    }
+      sc$state$spark_context <- if (spark_version(sc) >= "2.0") {
+
```
kevinykuo: We need to create a `SparkSession` anyway so it doesn't hurt performance to cache.

R/shell_connection.R
```diff
@@ -453,26 +453,27 @@ connection_is_open.spark_shell_connection <- function(sc) {
 
 #' @export
 spark_log.spark_shell_connection <- function(sc, n = 100, filter = NULL, ...) {
-  if (!is.null(sc$output_file) && file.exists(sc$output_file)) {
```
kevinykuo: Consolidate `spark_log.spark_shell_connection()` and `spark_log.spark_connection()`.

kevinykuo:
@javierluraschi tests are passing now, could you take another quick look?

R/spark_gateway.R
```diff
@@ -57,7 +57,7 @@ spark_gateway_connection <- function(master, config, gatewayInfo, gatewayAddress
   })
 
   # create the shell connection
-  sc <- structure(class = c("spark_connection", "spark_gateway_connection", "spark_shell_connection"), list(
```
mengxr: This change broke Databricks connection. Our test against master failed with:

~~~
Error in value[[3L]](cond) : 
  Failed during initialize_connection:Failed during initialize_connection: Unable to retrieve a spark_connection from object of class NULL
~~~

The error is from `spark_connection.default`. With the new refactor, we might need an implementation of `spark_connection` for Databricks.

@falaki @javierluraschi @kevinykuo Any recommendation for the fix?

PR #1683, opened 2018-09-14
broom for isotonic regression and aft survival regression

samuelmacedo83:
Hi @javierluraschi ,
I've done broom for isotonic regression and aft survival regression.
In tidy() for isotonic regression I just send a message "tidy() is not suported for Isotonic Regression", because in this model we don't have coefficients or something like that. if you have another idea, please let me now :)

here is a code that you can run

```
library(sparklyr)

sc <- spark_connect(master = "local", version = "2.2.1")
iris_tbl <- sdf_copy_to(sc, iris, name = "iris_tbl")

iris_tbl %>% 
  ml_isotonic_regression(Petal_Length ~ Petal_Width) %>% 
  tidy()

iris_tbl %>% 
  ml_isotonic_regression(Petal_Length ~ Petal_Width) %>% 
  augment()

iris_tbl %>% 
  ml_isotonic_regression(Petal_Length ~ Petal_Width) %>% 
  glance()

spark_disconnect(sc)

library(survival)

sc <- spark_connect(master = "local", version = "2.2.1")
ovarian_tbl <- sdf_copy_to(sc, ovarian, name = "ovarian_tbl", overwrite = TRUE)

ovarian_tbl %>% 
  ml_aft_survival_regression(futime ~ ecog_ps + rx, censor_col = "fustat") %>% 
  tidy()

ovarian_tbl %>% 
  ml_aft_survival_regression(futime ~ ecog_ps + rx + age + resid_ds, censor_col = "fustat") %>% 
  augment()

ovarian_tbl %>% 
  ml_aft_survival_regression(futime ~ ecog_ps + rx, censor_col = "fustat") %>% 
  glance()

spark_disconnect(sc)
```


javierluraschi:
@samuelmacedo83 I think we might want to use boundaries (as in `iso_res$model$boundaries()`) instead of `coefficients` for isotonic regression, @kevinykuo what do you think?

samuelmacedo83:
I think it is a good idea

samuelmacedo83:
@javierluraschi , I think that just print model$boundaries in tidy() could not be clean for the user. iris is a small data, I wondering when the user faces giga or terabytes and boundaries could be a very large vector.
```
library(sparklyr)
iris_tbl <- sdf_copy_to(sc, iris, name = "iris_tbl")
iso_reg <- iris_tbl %>% 
  ml_isotonic_regression(Petal_Length ~ Petal_Width)

iso_reg$model$boundaries
 [1] 0.1 0.1 0.2 0.2 0.2 0.2 0.2 0.4 0.4 0.4 0.4 0.6 1.0 1.0 1.0 1.1 1.1 1.1 1.2
[20] 1.2 1.2 1.3 1.3 1.3 1.3 1.4 1.4 1.4 1.4 1.5 1.5 1.5 1.6 1.6 1.8 1.8 1.8 2.0
[39] 2.0 2.3 2.3 2.5 2.5 2.5
```
One idea is print the boundaries with the frequencies, what do you think?

```
data.frame(boundaries = iso_reg$model$boundaries) %>% 
  group_by(boundaries) %>% 
  tally()
```

javierluraschi:
From looking at `boundaries` in the Scala API: https://spark.apache.org/docs/2.2.0/api/scala/index.html#org.apache.spark.ml.regression.IsotonicRegressionModel, they are defined as a Vector, which means that this is no longer distributed data across Spark.

So my guess is that the vector can get quite large but not large enough to not fit in memory, otherwise, it would have to be a reference to a Spark DataFrame not a Vector.

That said, as you mention, printing all those values could be more annoying than helpful.

So I think we want to do this,

We return `boundaries()` and also `predictions()` and provide a tibble to ensure that we bring all the data but only a subset is printed, something like:

```r
result <- tibble::data_frame(boundaries = iso_reg$model$boundaries(), predictions = iso_reg$model$predictions())
result
```

Then users could sample their very large data set, in our iris example, they can just run:

```r
plot(iris$Petal.Width, iris$Petal.Length)
```

Followed by the following (regardless of the dataset size):

```r
lines(result$boundaries, result$predictions)
```

isotonic-sparklyr

Since we are using a tibble, printing the tidy model should look like;

```
# A tibble: 44 x 2
   boundaries predictions
        <dbl>       <dbl>
 1        0.1        1.1 
 2        0.1        1.31
 3        0.2        1.31
 4        0.2        1.40
 5        0.2        1.4 
 6        0.2        1.5 
 7        0.2        1.52
 8        0.4        1.52
 9        0.4        1.6 
10        0.4        1.7 
# ... with 34 more rows
```

PR #1723, opened 2018-10-18
Properly clean JVM objects once deallocated from R

javierluraschi:


R/core_invoke.R
```diff
@@ -111,6 +111,10 @@ core_invoke_method <- function(sc, static, object, method, ...)
   backend <- core_invoke_socket(sc)
   connection_name <- core_invoke_socket_name(sc)
 
+  if (!identical(object, "Handler")) {
+    clearJobjs()
```
kevinykuo: Looking at the definition of `clearJobjs()` it seems like it's meant to be called upon `spark_disconnect()` instead of every `invoke()`, since it removes references on the R side for *all* JVM objects. However, it looks like we're not doing any explicit removal of objects from the JVM anywhere, even though we're marking things for deletion properly here https://github.com/rstudio/sparklyr/blob/94ab326b415a72c47ab2f7572e1c6a05be030313/R/core_jobj.R#L90. We probably want to invoke https://github.com/rstudio/sparklyr/blob/9f613f8a83713b3dbe770d728cb74b7e6c6a4eb3/java/spark-1.5.2/stream.scala#L44-L57 somewhere, like https://github.com/apache/spark/blob/5264164a67df498b73facae207eda12ee133be7d/R/pkg/R/backend.R#L75-L84. Haven't investigated what needs to be done exactly since there's some divergence of the backend code.

javierluraschi: Looks to me like we refactored this code into `clearJobjs()` which used to be called in every invoke. We could think of something different, but seems reasonable since `rm` wouldn't be called if the list of objects to clean is empty. See: https://github.com/apache/spark/blob/5264164a67df498b73facae207eda12ee133be7d/R/pkg/R/backend.R#L69-L84

What do you think? If it's easy to grow ML models in the JVM, I could verify this fix. However, when we collect data, we rather call a function that retrieves the data so no reference to the retrieved data is left in the JVM. This could be an issue in other workflows where we do keep JVM references in R, but I can't come up with any this time.

PR #1747, opened 2018-11-04
Broom for mlp

samuelmacedo83:
hi @javierluraschi ,

I've finished broom for ml_multilayer_perceptron_classifer(). For tidy() I created a data frame with the weight matrix by layer and for glance() I just put the number of nodes in each layer.
I add num_classes in new_ml_multilayer_perceptron_classification_model(), because broom_augment_supervised() need this. 
I didn't update NEWS.md yet, I'll put after the reviews.

```
sc <-  spark_connect(master = "local")

iris_tbl <- sdf_copy_to(sc, iris, name = "iris_tbl", overwrite = TRUE)

mlp_model <- iris_tbl %>%
  ml_multilayer_perceptron_classifier(Species ~ ., layers = c(4,3,2,3))

tidy(mlp_model)
augment(mlp_model)
glance(mlp_model)
```


tests/testthat/test-broom-pca.R
```diff
@@ -0,0 +1,56 @@
...
+test_that("pca.augment() works", {
+  test_requires_version("2.0.0")
+  sc <- testthat_spark_connection()
+  iris_tbl <- testthat_tbl("iris")
+
+
+  au1 <- iris_tbl %>%
+    dplyr::select(-Species) %>%
+    ml_pca(k = 3) %>%
+    augment(head(iris_tbl, 25)) %>%
```
javierluraschi: Would read better as `head(iris_tbl, 25) %>% augment() %>%`

R/tidiers_ml_multilayer_perceptron.R
```diff
@@ -0,0 +1,78 @@
...
+
+  dplyr::tibble(layers, weight_matrix)
+
+}
+
+
+
+
+
+
```
javierluraschi: Remove empty lines? We try to keep it to one line between functions. Hoping I can make some time to support `lintr` at some point to avoid style changes during code review.

NEWS.md
```diff
@@ -10,6 +11,27 @@
   not yet implemented in Spark.
   
 - Support for collecting NA values from logical columns (#1729).
+||||||| merged common ancestors
+<<<<<<<<< Temporary merge branch 1
```
javierluraschi: I believe you mentioned you are planning to fix this merge conflicts, just adding this note to make sure we don't forget.

javierluraschi:
LGTM! @kevinykuo anything else to add?

R/tidiers_pca.R
```diff
@@ -0,0 +1,49 @@
+#' Tidying methods for Spark ML Principal Component Analysis
```
javierluraschi: Strictly speaking, we don't need this file for now, let me send a mail to you and Kevin to figure out how to tackle the remaining models.

kevinykuo:
Looks great, couple comments:
1. Can we replace the `for` loops with `map()` calls to be consistent with the style of the codebase?
2. For the glance method, I think we should make it clear in the column names that these are number of units, and also put the hidden layers into a vector so we don't end up with really wide data frames for deep networks.

javierluraschi:
I don't think we would end up with very wide data, as in, very deep NN are like 7-20 layers? I would be concerned if we had thousands, etc. Nested vectors are still harder to deal with than nested data, since values are not previewed while printing and the tools to unnest data are newer, so my preference would be to keep this matrix flat.

PR #1753, opened 2018-11-06
Support for tables with periods

javierluraschi:
```r
sensors <- data.frame(
    sensor = c("s.ensor@1", "s.ensor@1","s.ensor@1","s.ensor@2","s.ensor@2","s.ensor@2"),
    time = c(1,2,3,1,2,3),
    metric = c(1,2,3,10,20,30)
)

sensors_tbl <- sdf_copy_to(sc, sensors, overwrite = T)

sensors_tbl %>%
  sdf_pivot(time ~ sensor, fun.aggregate = c("avg", "metric")) %>%
  select(`s.ensor@1`)
```

```
# Source: spark<?> [?? x 1]
  `s.ensor@1`
*       <dbl>
1           1
2           3
3           2
```

Consider using instead [dbplyr::in_schema(database, table)](https://www.rdocumentation.org/packages/dbplyr/versions/1.2.2/topics/in_schema) instead.

This can also be reverted to match previous versions through:

```r
sc <- spark_connect(master = "local", config = list(sparklyr.dplyr.period.splits = T))
sensors_tbl <- sdf_copy_to(sc, sensors, overwrite = T)

sensors_tbl %>%
  sdf_pivot(time ~ sensor, fun.aggregate = c("avg", "metric")) %>%
  select(`s.ensor@1`) %>% dbplyr::sql_render()
```
```
<SQL> SELECT `s`.`ensor@1`
FROM `sparklyr_tmp_f23019c81436`
```

Current workaround:

```r
sensors_tbl %>%
  mutate(sensor = regexp_replace(sensor, "\\\\.", "_")) %>%
  sdf_pivot(time ~ sensor, fun.aggregate = c("avg", "metric")) %>%
  select(`s_ensor@1`)
```
```
# Source: spark<?> [?? x 1]
  `s_ensor@1`
*       <dbl>
1           1
2           3
3           2
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,9 +1,8 @@
 # Sparklyr 0.9.3 (unreleased)
 
-- Tables with periods are now supported; however, this is a 
-  **breaking change** for users referencing tables as `database.table` 
-  since now they will have to set the `sparklyr.dplyr.period.splits`
-  to `TRUE`.
+- Tables with periods are now supported; for `database.name` use instead
+  `dbplyr::in_schema()` or, for backwards compatibility set
+  `sparklyr.dplyr.period.splits` to `TRUE`.
 
 - Fix requirement to specify `SPARK_HOME_VERSION` when `version`
   parameter is set in `spark_connect()`.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # Sparklyr 0.9.3 (unreleased)
 
+- Tables with periods are now supported; however, this is a 
+  **breaking change** for users referencing tables as `database.table` 
+  since now they will have to set the `sparklyr.dplyr.period.splits`
+  to `TRUE`.
+
 - Fix requirement to specify `SPARK_HOME_VERSION` when `version`
   parameter is set in `spark_connect()`.
 
--- a/R/config_settings.R
+++ b/R/config_settings.R
@@ -21,6 +21,7 @@ spark_config_settings <- function() {
     sparklyr.connect.packages = "Spark packages to include when connecting to Spark.",
     sparklyr.connect.sparksubmit = "Command executed instead of spark-submit when connecting.",
     sparklyr.connect.timeout = "Total seconds before giving up connecting to the sparklyr gateway while initializing.",
+    sparklyr.dplyr.period.splits = "Should 'dplyr' split column names into database and table?",
     sparklyr.gateway.address = "The address of the driver machine.",
     sparklyr.gateway.config.retries = "Number of retries to retrieve port and address from config, useful when using functions to query port or address in kubernetes.",
     sparklyr.gateway.interval = "Total of seconds sparkyr wil check on a gateway connection.",
--- a/R/dplyr_spark_connection.R
+++ b/R/dplyr_spark_connection.R
@@ -4,7 +4,7 @@
 sql_escape_ident.spark_connection <- function(con, x) {
   # Assuming it might include database name like: `dbname.tableName`
   if (length(x) == 1)
-    tbl_quote_name(x)
+    tbl_quote_name(sc, x)
   else
     dbplyr::sql_quote(x, '`')
 }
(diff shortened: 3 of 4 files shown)


javierluraschi:
@dkincaid fair enough, see https://github.com/rstudio/sparklyr/pull/1776. However, would it be feasible for your team to eventually deprecate this behavior, we really want to encourage the right solution here which is `dbplyr::in_schema()`.

dkincaid:
If we aren't doing something the right way, I'm happy to change, but I don't understand what the right way is right now. If I have a table called `animals` where one column is a struct. Let's say the column name is `species` and it is a struct with elements named `code` and `display`. Right now we can do something like this:
```
animals %>% filter(species.code == "448771007")
```
to filter the table to only records with a certain species code. What should we be doing?

javierluraschi:
@dkincaid so this change actually makes your code work, if I turn off this fix this happens:

```r
library(sparklyr)
library(dplyr)

sc <- spark_connect(
  master = "local",
  config = list(
    sparklyr.sanitize.column.names = FALSE,
    sparklyr.dplyr.period.splits = TRUE
  )
)

animals <- data.frame(species.code = "448771007", something = "a")

animals_tbl <- copy_to(sc, animals)
animals_tbl
```
```r
# Source: spark<animals> [?? x 2]
  species.code something
* <chr>        <chr>    
1 448771007    a  
```
```r
animals_tbl %>% filter(species.code == "448771007")
```
```
 Error: org.apache.spark.sql.AnalysisException: cannot resolve '`species.code`' given input columns: [animals.species.code, animals.something]; line 3 pos 7;
'Project [*]
+- 'Filter ('species.code = 448771007)
   +- SubqueryAlias animals
      +- LogicalRDD [species.code#8, something#9], false
```

However, with this fix:

```r
spark_disconnect(sc)
sc <- spark_connect(
  master = "local",
  config = list(
    sparklyr.sanitize.column.names = FALSE
  )
)
animals_tbl <- copy_to(sc, animals)
animals_tbl %>% filter(species.code == "448771007")
```
```
# Source: spark<?> [?? x 2]
  species.code something
* <chr>        <chr>    
1 448771007    a     
```

So if anything, I think you are arguing that we should make this change. Right?

To give you more context, this is to fix cases where users do:

```r
tbl(sc, "database.animals") %>% count()
```

Where `database` is a different database than the current one, and `animals` is the table you are accessing. I believe this functionality is not used very much anyways. That code would have to change to:

```r
```r
tbl(sc, dbplyr::in_schema("database", "animals")) %>% count()
```
```

Do you have code you can share were you've seen this broken with this fix?

**Comment by javierluraschi** [comment]
@dkincaid Wait, you are talking about structured fields... yes... that's a very valid use case. Give me a sec...

**Comment by mattpollock** [comment]
I think a closer reprex would be

```r
iris_spk <- copy_to(sc, iris)
iris_nst %>% sparklyr.nested::sdf_nest(Species, Sepal_Width)
iris_nst %>% filter(data.Species == "setosa")
```

dkincaid:
Yes, let me get the code. Because this change actually breaks this. In my example, the column name isn't `species.code`. The column name is `species`. The column type is a Struct with properties `code` and `display`. Think of a Java object `Species` with two properties `String code` and `String display`. It's natural to refer to the code of an instance of a Species object as `species.code` and this is in fact how the SparkSQL works in this case. I can write

```
tbl(sc, sql("SELECT * FROM animals WHERE `species`.`code`= '448771007'"))
```
(I would not actually use the backticks here since they are unnecessary, but they show why this change breaks this)
but with this change the sql this is getting created looks like this:
```
tbl(sc, sql("SELECT * FROM animals WHERE `species.code`= '448771007'"))
```
which throws an error that the column `species.code` isn't found.

I can pull together an example and provide it a little bit later.

dkincaid:
Yes, your example is close to what I was going to put together for you.

javierluraschi:
Makes sense, so sorry about this. The CRAN version of `sparklyr` does not contain this fix, fixing now `devel`, thanks @dkincaid for bringing this up!

javierluraschi:
We certainly appreciate you running the `devel` version, it really helps us find issues earlier and fix them. Would you mind sending me a mail directly yo `javier` at `rstudio.com` if something breaks compatibility in `devel`? Keeping `sparklyr` stable is certainly more important than any new feature we introduce.

Issue #1453, opened 2018-04-27
ML model object member semantics


kevinykuo:
Currently, we're calling `invoke()` to compute all members on object instantiation in R. This can be expensive. Instead consider returning closures or not return anything at all then extract when needed using helper functions.

PR #1778, opened 2018-11-20
[WIP] Developer API for ML pipeline extensions

kevinykuo:
Changes in this PR:
- Expose a developer API to create sparklyr extensions that wrap Spark packages written in Scala that implement Pipelines (i.e. Estimators and Transformers.) A developer guide will be published after merge.
- Internal refactoring to support the developer API. Most of this should have mostly no effect on current users, with the exceptions below.

Potentially breaking changes:

- Class hierarchies have been modified to better match the Spark style of inheritance, e.g. 
```
> class(ml_logistic_regression(sc))
[1] "ml_logistic_regression"      "ml_probabilistic_classifier"
[3] "ml_classifier"               "ml_predictor"               
[5] "ml_estimator"                "ml_pipeline_stage"  
```
  - All `lazy val` and `def` members of model objects are now properly closures  (https://github.com/rstudio/sparklyr/issues/1453).


R/core_serialize.R
```diff
@@ -17,8 +17,21 @@
 # jobj -> Object, where jobj is an object created in the backend
 # nolint end
 
+get_type <- function(object, types = NULL) {
```
kevinykuo: @javierluraschi Could you review this change right here? Basically it's to accommodate things like
```
> int_1 <- forge::cast_scalar_integer(1)
> int_1
[1] 1
> class(int_1)
[1] "forge_stamped" "integer"   
```

PR #1783, opened 2018-11-26
Improve type error

MichaelChirico:
I was met with this confusing error and spent some time debugging it:

> Unsupported type for serialization glue

I think I would have realized what was going on right away if the message was instead:

> Unsupported type 'glue' for serialization

I also noticed some usage of the `stop(paste(` form of building error messages, which is discouraged where it can be avoided (it's why `stop` has `...` argument). This is a separate commit as it's more cosmetic and can be more easily reversed.

I wasn't sure if I should add tests; let me know.


javierluraschi:
Thanks @MichaelChirico, looks good to me.

PR #1785, opened 2018-11-26
Dereference symlinks when creating package bundle

awblocker:
Currently, `spark_apply_bundle` will bundle symlinks, such as those used by `packrat`, as-is. This leads to errors when the resulting bundle is deployed to workers with symlinks intact. Adding `-h` to the `tar` command forces `tar` to dereference symlinks, resulting in the actual package files being bundled and resolving this issue.


javierluraschi:
This is a great catch, will add this to the `NEWS` to make folks aware of this change.

PR #1786, opened 2018-11-27
Modifications in ml_lda() and broom

samuelmacedo83:
Hi @javierluraschi ,

I added formula and tokenizer-count-vectorizer pipeline to ml_lda(). I also made broom for it.

Here is a code you can run:

```
library(janeaustenr)
sc <-  spark_connect(master = "local")

lines_tbl <- sdf_copy_to(sc,
                         austen_books()[c(1:30),],
                         name = "lines_tbl",
                         overwrite = TRUE)

lda_model <- lines_tbl %>%
  ml_lda(~text, k = 4)

tidy(lda_model)
augment(lda_model)
glance(lda_model)
```


kevinykuo:
Thanks! I'll take a look in the next day or two!

R/ml_model_lda.R
```diff
@@ -0,0 +1,13 @@
...
+                                dataset, formula, feature_names,
+                                call) {
+
+  vocabulary <- pipeline_model$stages[[2]]$vocabulary
+
+  new_ml_model_clustering(
+    pipeline, pipeline_model,
+    model, dataset, formula,
+    vocabulary = vocabulary,
+    subclass = "ml_lda_model"
```
kevinykuo: This should be `"ml_model_lda"`, since `ml_lda_model` is the class for the pipeline stage; see `new_ml_lda_model()`.

R/ml_model_utils.R
```diff
@@ -92,6 +100,11 @@ ml_generate_ml_model <- function(
   else
     ml_uid(predictor)
 
+# ml_features_names_metadata() just verifies the first stage (ml_stage(pipeline_model, 1)
+# and ml_lda() has more than one preprocessor
+if (grepl("lda", predictor$uid)) {
```
kevinykuo: It's better practice to check for `inherits(predictor, "ml_lda")` here since `uid` could be arbitrary (even though it's going to use the default in this case.)

R/tidiers_ml_lda.R
```diff
@@ -0,0 +1,59 @@
...
+#' @param ... extra arguments (not used.)
+#' @name ml_lda_tidiers
+NULL
+
+#' @rdname ml_lda_tidiers
+#' @export
+tidy.ml_lda_model <- function(x,
+                              ...){
+
+ doc_concentration <- sparklyr::invoke(x$model$estimated_doc_concentration(),
```
kevinykuo: Let's put some line breaks in these functions calls, see https://style.tidyverse.org/syntax.html#long-lines.

R/tidiers_ml_lda.R
```diff
@@ -0,0 +1,59 @@
...
+
+ cbind(describe_topics,
+       doc_concentrarion = doc_concentration) %>%
+    dplyr::as_tibble()
+}
+
+#' @rdname ml_lda_tidiers
+#'
+#' @export
+augment.ml_lda_model <- function(x,
```
kevinykuo: We don't need to break here. Also, space between `)` and `{`.

tests/testthat/test-broom-lda.R
```diff
@@ -0,0 +1,68 @@
...
+                         name = "lines_tbl",
+                         overwrite = TRUE)
+
+  td1 <- lines_tbl %>%
+    ml_lda(~text, k = 3) %>%
+    tidy()
+
+  check_tidy(td1, exp.row = 3, exp.col = 4,
+             exp.names = c("topic", "termIndices",
+                           "termWeights", "doc_concentrarion"))
```
kevinykuo: Is there precedence in these names? Could we just use snake case? Also, typo in `doc_concentration`.

kevinykuo:
Actually here are some examples for topicmodels + broom https://cran.r-project.org/web/packages/tidytext/vignettes/topic_modeling.html. I'd be inclined to follow whatever convention they've set there.

samuelmacedo83:
hi  @kevinykuo, 
thanks for the feedback and the [reference](https://cran.r-project.org/web/packages/tidytext/vignettes/topic_modeling.html).

I'll make the changes

kevinykuo:
Okay awesome, there's some breaking changes I'm working on in https://github.com/rstudio/sparklyr/pull/1778 so I'll reconcile this coming week.

kevinykuo:
@samuelmacedo83 FYI I haven't forgotten about this one. The dev API is taking longer than expected, but hopefully should have everything wrap up in the next few days and I'll assist with modifying this code to work with the new internals. Thanks for the patience.

samuelmacedo83:
HI @kevinykuo , glad to hear from you. 
Don't worry, I know you are busy with the API. When you finish it we work on this branch.
Thank for answer me :)

kevinykuo:
OK here are the relevant internal API changes.

- Formula standardization is now done via `ml_standardize_formula()`, e.g. https://github.com/rstudio/sparklyr/blob/master/R/ml_clustering_kmeans.R#L88.
- The `ml_model` construction now uses a `ml_construct_model_*` helper which takes the specific `new_ml_model_*` constructor for that class, e.g. https://github.com/rstudio/sparklyr/blob/master/R/ml_clustering_kmeans.R#L109-L115. You can look at the signatures for these functions to get a better picture.
- The `new_ml_model_*` constructor for the class should call the appropriate parent constructor. Probably one of the functions defined here https://github.com/rstudio/sparklyr/blob/master/R/ml_model_constructors.R. Hopefully LDA should be able to use the `ml_model_clustering` constructor, if not, you may need to inherit from `ml_model` directly. If that doesn't provide enough flexibility, let me know and we'll figure out how to do it.

samuelmacedo83:
Hi @kevinykuo, thank you to give a summary of the modifications, it helped me a lot :)

I've made the update in LDA and broom-LDA according to the API changes. 
I had to put an if statement for LDA inside new_ml_model_clustering() because LDA uses more than one preprocessor and ml_feature_names_metadate() considers just one. Let me know if you have another idea to do that.

PR #1797, opened 2018-12-06
Unique SqlContext per databricks connection

MrBago:
In databricks we create a unique SqlContext (HiveContext) for each notebook or rstudio session even though all users share the same spark context. Because sparklyr uses getOrCreate to get the SqlContext for each user, our users end up sharing a SqlContext. This has causes unexpected behavior and errors for our sparklyr users.

This PR allows databricks connections to get the right SqlContext upfront, resolving this issue.

### DIFF
--- a/R/databricks_connection.R
+++ b/R/databricks_connection.R
@@ -1,7 +1,7 @@
 databricks_connection <- function(config, extensions) {
   tryCatch({
     callSparkR <- get("callJStatic", envir = asNamespace("SparkR"))
-    # In Databricks notebooks, this variable is in the default namespace
+    # In Databricks environments (notebooks & rstudio) DATABRICKS_GUID is in the default namespace
     guid <- get("DATABRICKS_GUID", envir = .GlobalEnv)
     gatewayPort <- as.numeric(callSparkR("com.databricks.backend.daemon.driver.RDriverLocal",
                                          "startSparklyr",
@@ -39,8 +39,10 @@ new_databricks_connection <- function(scon, guid) {
     scon,
     class = "databricks_connection"
   )
-  rDriverLocal <- "com.databricks.backend.daemon.driver.RDriverLocal"
-  hive_context <- invoke_static(sc, rDriverLocal, "getDriver", guid) %>% invoke("sqlContext")
+  # In databricks, sparklyr should use the SqlContext associated with the RDriverLocal instance for
+  # this guid.
+  r_driver_local <- "com.databricks.backend.daemon.driver.RDriverLocal"
+  hive_context <- invoke_static(sc, r_driver_local, "getDriver", guid) %>% invoke("sqlContext")
   sc$state$hive_context <- hive_context
   sc
 }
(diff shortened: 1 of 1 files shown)


MrBago:
@javierluraschi @falaki would you mind taking a look?

R/databricks_connection.R
```diff
@@ -1,10 +1,11 @@
 databricks_connection <- function(config, extensions) {
   tryCatch({
     callSparkR <- get("callJStatic", envir = asNamespace("SparkR"))
+    # In Databricks notebooks, this variable is in the default namespace
```
falaki: The comment can be more general. On Databricks (notebooks or RStudio) `DATABRICKS_GUID` is available in the global namespace.

R/databricks_connection.R
```diff
@@ -32,9 +34,13 @@ spark_version.databricks_connection <- function(sc) {
...
 }
 
-new_databricks_connection <- function(scon) {
-  new_spark_gateway_connection(
+new_databricks_connection <- function(scon, guid) {
+  sc <- new_spark_gateway_connection(
     scon,
     subclass = "databricks_connection"
   )
+  rDriverLocal <- "com.databricks.backend.daemon.driver.RDriverLocal"
```
falaki: Nit: I think sparklyr uses snake_case naming.

R/databricks_connection.R
```diff
@@ -32,9 +34,13 @@ spark_version.databricks_connection <- function(sc) {
...
+  hive_context <- invoke_static(sc, rDriverLocal, "getDriver", guid) %>% invoke("sqlContext")
+  sc$state$hive_context <- hive_context
```
falaki: Please add  a short in-line comment on why we are doing this.

javierluraschi:
I don't have further feedback, we can merge after addressing existing comments.

MrBago:
@javierluraschi thanks for taking a look. After I made this PR I was looking trough the codebase a bit and realized that maybe we need to move the fix into here, https://github.com/rstudio/sparklyr/blob/2b6554fbcea153abcf185af46cf4297e1c69b0b2/R/shell_connection.R#L597-L616.

Specifically I'm worried about line 615, it seems like with the current solution, the R side and backend might have handles to different sqlContexts. I'm new to sparklyr so I'm not sure if I've got the flow right or if it matters. @falaki?

Also, does anyone know what's going on with the CI. @kevinykuo I noticed you've switched from using openjdk to oracle for your PR.

javierluraschi:
@MrBago I have limited knowledge as to how `sparklyr` is initialized in Databricks; however, from that particular code, it seems to be the case that `sparklyr` won't re-initialize the `sqlContext` if it got already initialized in the Databricks override. See use of `%||%` operator in [#L618](https://github.com/rstudio/sparklyr/blob/2b6554fbcea153abcf185af46cf4297e1c69b0b2/R/shell_connection.R#L618).

You can try rebasing to commit 1459234c5d29316872e71b0880fb0329695ea6f4 to get the tests to pass.

falaki:
LGTM

PR #1871, opened 2019-02-01
Broom for als

samuelmacedo83:
HI @javierluraschi and @kevinykuo ,

I've finished broom for als model. I didn't code for ml_als_factorization() because it'll be deprecated.

Here is a code that you can run:

```
library(sparklyr)

movies <- data.frame(user   = c(1, 2, 0, 1, 2, 0),
                     item   = c(1, 1, 1, 2, 2, 0),
                     rating = c(3, 1, 2, 4, 5, 4))
movies_tbl <- sdf_copy_to(sc, movies, name = "movies_tbl", overwrite = TRUE)
model <- ml_als(movies_tbl)

tidy(model)
augment(model)
glance(model)
```


kevinykuo:
I don't think we want to implement broom methods for any class that doesn't inherit from `ml_model` for consistency. Could we introduce a formula interface for `ml_als.tbl_spark()` somehow so we can signal that we want to create an `ml_model_als`?

Feel free to open a separate PR to fix the bug with `find_constructor()` btw, thanks for the catch!

samuelmacedo83:
Ok, I'll put the fix in ```find_constructor()``` in another PR.

I noticed that als just uses ```ml_fit()``` instead of ```ml_construct_model_```.
We could create ```ml_construct_model_recommendation()``` inside of  ```ml_als.tbl_spark()``` and 
create a ```new_ml_model_als``` constructor. 

Something like that:
```
new_ml_model_als <- function(pipeline_model, formula, dataset,
                                     features_col) {

  new_ml_model_recommendation(
    pipeline_model = pipeline_model,
    formula = formula,
    dataset = dataset,
    features_col = features_col,
    class = "ml_model_als"
  )
}
```
what do you think?

kevinykuo:
But how would we know whether to just call `ml_fit()` or `ml_construct_model_recommendation()`? For the other algorithms we have something like
```
  if (is.null(formula)) {
    stage %>%
      ml_fit(x)
  } else {
    ml_construct_model_supervised(
      new_ml_model_linear_regression,
      predictor = stage,
      formula = formula,
      dataset = x,
      features_col = features_col,
      label_col = label_col
    )
  }
```
Basically, we need to have a sensible way for users to provide a formula to signal they want an `ml_model_als` back instead of an `ml_als_model`.

samuelmacedo83:
I added the formula interface for ml_als(). 
I also did some modifications in ml_recommend(), and now it also works for ml_model_als.

Here is an example:

```
library(sparklyr)

movies <- data.frame(
  user   = c(1, 2, 0, 1, 2, 0),
  item   = c(1, 1, 1, 2, 2, 0),
  rating = c(3, 1, 2, 4, 5, 4)
)
movies_tbl <- sdf_copy_to(sc, movies)

model <- ml_als(movies_tbl, rating ~ user + item)

tidy(model)
augment(model)
glance(model)
ml_recommend(model, type = "item", 1)

```

R/ml_model_als.R
```diff
@@ -0,0 +1,16 @@
+new_ml_model_als <- function(pipeline_model,
+                             formula,
+                             dataset,
+                             label_col,
+                             features_col) {
+ m <- new_ml_model_regression(
```
kevinykuo: We shouldn't inherit ml_model_regression here since it's not a regression model and there are no `label_col`, `features_col`, etc.

samuelmacedo83: is it better to create a new_model_recommendation?

kevinykuo: yeah I think that's the right thing to do

samuelmacedo83: Ok, I'll work on it

R/ml_model_constructors.R
```diff
@@ -165,6 +165,20 @@ new_ml_model_clustering <- function(pipeline_model, formula, dataset,
   )
 }
 
+#' @export
```
kevinykuo: Let's not export these constructors since I don't see any other packages using them.

R/ml_recommendation_als.R
```diff
@@ -263,7 +280,16 @@ utils::globalVariables("explode")
 #'
 #' @export
 ml_recommend <- function(model, type = c("items", "users"), n = 1) {
-  if (spark_version(spark_connection(model)) < "2.2.0")
+
+  if(inherits(model, "ml_model_als")){
```
kevinykuo: Could we refactor this to check for the version and throw the error just once? e.g. `version <- if (inherits ml model) {} else {}` then check `version`.

tests/testthat/test-broom-als.R
```diff
@@ -0,0 +1,68 @@
...
+
+test_that("als.augment() works", {
+  test_requires_version("2.0.0")
+  sc <- testthat_spark_connection()
+
+  movies <- data.frame(user   = c(1, 2, 0, 1, 2, 0),
+                       item   = c(1, 1, 1, 2, 2, 0),
+                       rating = c(3, 1, 2, 4, 5, 4))
+
+  movies_tbl <- sdf_copy_to(sc,
```
kevinykuo: Let's use `copy_to()` (which is now re-exported) throughout since we'll be deprecating `sdf_copy_to()`.

kevinykuo:
Will follow up with a separate PR

PR #1883, opened 2019-02-06
Improve LDA formula interface behavior

kevinykuo:
- Use temporary names for feature transformer columns to avoid clashes.
- Use regex tokenizer instead of tokenizer, which is more robust (e.g. splits on `\\s+` instead of ` `).
- Add a stop words remover stage.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
-Version: 0.9.9002
+Version: 0.9.9003
 Authors@R: c(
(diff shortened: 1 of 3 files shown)


javierluraschi:
In case this helps, looks like tests need to be updated:

```
Context: broom-lda
lda.tidy() works: failure: nrow(o) not equal to `exp.row`.
1/1 mismatches
[1] 18 - 27 == -9 
lda.tidy() works: failure: td1$beta[1:3] not equal to c(0.879, 0.946, 1.15).
```

kevinykuo:
Yep, @samuelmacedo83 has https://github.com/rstudio/sparklyr/pull/1886 fixing these, closing.

PR #1915, opened 2019-02-19
Fix for Databricks when connecting with Spark 2.4

javierluraschi:
Databricks manually loads the `sparklyr` JARs, some are missing since we simplified the JARs we ship, this PR adds the missing JARs when connecting in Databricks.

screen shot 2019-02-18 at 5 57 20 pm

Also fixing `hive_context()` issue where databricks passes a SQLContext instead of HiveContext. A few bonus fixes to properly support parameterized queries when using `DBI`.

screen shot 2019-02-19 at 12 47 06 pm


javierluraschi:
Intermittent `r-appveyor` issue unrelated to this change, merging.

PR #1925, opened 2019-02-22
Add support for sparklyr extensions in Livy

javierluraschi:
This PR adds support for Livy extensions, extensions will have to support a new `web_jars` parameter with a location describing where JARs can be downloaded from.

See for instance https://github.com/javierluraschi/geospark/commit/abc70ca18906e479944e8685243074652c39afb1.

Extensions will have to support the `catalog` parameter in `spark_dependency` which should default to an accessible location to download the extensions JARs from.

Users can also create their own private catalogs and override with the `sparklyr.extensions.catalog` setting.


kevinykuo:
Could you also test to see if this works with sparkxgb?

Issue #1956, opened 2019-03-21
sprark_apply spark.r.libpaths config max 1 path


mattpollock:
If I attempt to run spark with more than one shared library path (e.g., a user path and a system path) things don't work.

This fails
```
library(sparklyr)

spark_disconnect_all()
config <- list(spark.r.libpaths = c("/path/one", "/path/two"))
sc <- spark_connect(master = "local", config = config)
```
In local mode I get the following error
```
Error in force(code) : 
  Failed during initialize_connection: java.lang.IllegalArgumentException: length and stream do not match
	at sparklyr.Serializer.readStringBytes(serializer.scala:77)
	at sparklyr.Serializer.readString(serializer.scala:84)
	at sparklyr.Serializer.readTypedObject(serializer.scala:37)
       ....
```

against my yarn cluster I get
```
Error in force(code) : 
  Failed during initialize_connection: Result 12 must be a single string, not a character vector of length 2
    Log: /tmp/Rtmpntdl41/file47b7587451a8_spark.log
```

This is fine
```
spark_disconnect_all()
config <- list(spark.r.libpaths = "/path/two")
sc <- spark_connect(master = "local", config = config)
```

harryprince:
```
config <- list(spark.r.libpaths = c("/path/one;/path/two"))
```

this means set LIBPATHS environment variable as "/path/one;/path/two"

mattpollock:
@harryprince I'm not quite following. I had not tried to collapse the paths with a `;` delimiter, but tried that this morning without success. Using 
```
config <- list(spark.r.libpaths = c("/path/one;/path/two"))
```
in the configuration results in calling
```
.libPaths("/path/one;/path/two")
```
in the R process on the spark executor, which fails and I'm left with only the default package library paths on the worker node.

mattpollock:
It seems to me the this could be solved either by choosing a delimiter and `strsplit` (or similar) on the worker side to split the paths into a character vector or by enabling character vectors in the configuration. Not which which of these @javierluraschi would prefer.

mattpollock:
Proposed solution (in the PR) is to use a comma separator, so users would use something like
```r
config <- list(spark.r.libpaths = c("/path/one,/path/two"))
```

PR #1958, opened 2019-03-26
Bugfix/spark r libpaths

mattpollock:
for #1956 it looks like it will be easier to use a delimiter in the `spark.r.libpaths` configuration than to change the `purrr` pipeline that complains when it meets a vector of strings (in `apply_config`).


R/worker_apply.R
```diff
@@ -45,7 +45,10 @@ spark_worker_init_packages <- function(sc, context) {
   else {
     spark_env <- worker_invoke_static(sc, "org.apache.spark.SparkEnv", "get")
     spark_libpaths <- worker_invoke(worker_invoke(spark_env, "conf"), "get", "spark.r.libpaths", NULL)
-    if (!is.null(spark_libpaths)) .libPaths(spark_libpaths)
+    if (!is.null(spark_libpaths)) {
+      spark_libpaths <- unlist(strsplit(spark_libpaths, split = ","))
```
mattpollock: I specifically used base R functions here since I think we can make no assumptions as of when this runs that any non-base packages will be available.

javierluraschi:
Thanks @mattpollock, good catch!

R/worker_apply.R
javierluraschi: That is correct... all the `worker_` and `core_` files can't depend on any packages since they are used in the worker nodes to load packages in `spark_apply()`.

mattpollock:
Any idea why updating the NEWS broke the `ft_string_indexer` test? All checks passed as of 20f406e

https://travis-ci.org/rstudio/sparklyr/jobs/511725626#L1727

javierluraschi:
Unrelated, package changed in CRAN that broke code to handle previous interfaces in `ft_string_indexer`. Fixed now, merging.

PR #1967, opened 2019-04-01
Update test-dplyr.R

benmwhite:
Possible typo in the test for the `filter()` implementation. Original version compared two identical filtering expressions involving the local data frame `iris` with no reference to `iris_tbl` in Spark. Using `nrow()` instead of the Species value for the comparison to avoid converting between factor/string and data.frame/tibble within the test, output should be "1" in both cases.


javierluraschi:
Following up with https://github.com/rstudio/sparklyr/pull/2069

PR #1988, opened 2019-04-24
Rename sdf_partition() to sdf_random_split()

kevinykuo:
"partition" means something special in spark and the old name can cause confusion.


kevinykuo:
Merging and opened https://github.com/rstudio/sparklyr/issues/1996 to track failure with 1.6.

PR #2012, opened 2019-05-17
Support for Spark 2.4.3 and 2.3.3

javierluraschi:


h-vetinari:
Two questions (quite urgent for my case - can't use sparklyr ATM due to a CDH upgrade):
- does this also work for hadoop 3.0? (the segments in ` inst/extdata/versions.json` only go up to hadoop 2.7)
- would it be possible to release a minor release with this, please please please? (yes, even if there's just been one commit since 1.0.1)

h-vetinari:
I have a question: does this also work for hadoop 3.0?

The segments in ` inst/extdata/versions.json` only go up to hadoop 2.7, for all spark versions...

PR #2019, opened 2019-05-24
Fix Arrow devel builds on Travis

nealrichardson:
Thanks to @kou for the fix. You can see the passing build [here](https://travis-ci.org/nealrichardson/sparklyr/builds/536416026).

This also pins the version (changeset) for the arrow build because the R package is broken on the current head of master. Once that issue is resolved, I'll make a PR to remove the pin.


javierluraschi:
Nice! Thanks Neal! We can revert once upstream is fixed with https://github.com/rstudio/sparklyr/issues/2027

Issue #2016, opened 2019-05-21
Connecting to Spark through Livy failed with "File name too long"


lgongmsft:
Tried to connect to spark through livy but got an error saying "File name too long".
Does anybody have any idea?  Here is my simple repro script
```r
library(sparklyr)

config <- livy_config(username="<username>", password="<password>")

config$sparklyr.verbose <- TRUE
config$sparklyr.log.console <- TRUE
config$sparklyr.log.invoke <- TRUE
config$sparklyr.connect.timeout <- 60 * 5

sc <- spark_connect(master = "<address>", method = "livy", config = config)
spark_disconnect(sc)
```
and here is what I got:
```r
...
Loading spark-1.5.2/mlutils.scala
Error in value[[3L]](cond) :
  Failed to initialize livy connection: Failed to load mlutils.scala: Failed to execute Livy statement with error: <console>:134: error: File name too long
Traceback: This can happen on some encrypted or legacy file systems.  Please see SI-3623 for more details.
           .setStages(explodePipeline(pipeline) ++ stages.flatMap(explodePipeline))
                                                                  ^
<console>:153: error: File name too long
This can happen on some encrypted or legacy file systems.  Please see SI-3623 for more details.
           f(pipeline) filterNot {_.isInstanceOf[Pipeline]}
                                                ^
<console>:170: error: File name too long
This can happen on some encrypted or legacy file systems.  Please see SI-3623 for more details.
             pair => (pair.param.name -> pair.value)
                  ^
<console>:162: error: File name too long
This can happen on some encrypted or legacy file systems.  Please see SI-3623 f
Calls: spark_connect ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>
Execution halted
```
Here is my sessionInfo:
```r
> library(sparklyr)
> sessionInfo()
R version 3.5.2 (2018-12-20)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 16.04.6 LTS

Matrix products: default
BLAS: /opt/microsoft/ropen/3.5.2/lib64/R/lib/libRblas.so
LAPACK: /opt/microsoft/ropen/3.5.2/lib64/R/lib/libRlapack.so

locale:
[1] C

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

other attached packages:
[1] sparklyr_0.9.4       RevoUtils_11.0.2     RevoUtilsMath_11.0.0

loaded via a namespace (and not attached):
[...]

PR #2020, opened 2019-05-25
Fix version parameter for livy in docs

javierluraschi:
Fix for https://github.com/rstudio/sparklyr/issues/2016


javierluraschi:
Thanks for the review @lgongmsft!

PR #2030, opened 2019-06-03
Remove arrow changeset pin (fixes #2027)

### LINKED ISSUE #2027: Revert temporary arrow-devel workaround
See https://github.com/rstudio/sparklyr/pull/2019/files#diff-784b2c3739f6ce3e3c9ab8e0b049df43

nealrichardson:
Romain's arrow fix (https://issues.apache.org/jira/browse/ARROW-5361) has been merged, so this can be removed.

### DIFF
--- a/ci/arrow-build.sh
+++ b/ci/arrow-build.sh
@@ -9,9 +9,6 @@ sudo apt install -y -V autoconf
 if [[ $ARROW_VERSION == "devel" ]]; then
   git clone https://github.com/apache/arrow
   cd arrow/cpp
-  # arrow@master has a broken R package, pending https://issues.apache.org/jira/browse/ARROW-5361
-  # Remove this version pin after that issue is resolved
-  git checkout 1ed608aa94f0d22cfa69c81687006b4ebaefa258
 else
   wget https://arrowlib.rstudio.com/dist/arrow/arrow-$ARROW_VERSION/apache-arrow-$ARROW_VERSION.tar.gz
   tar -xvzf apache-arrow-$ARROW_VERSION.tar.gz


javierluraschi:
Thanks @nealrichardson! We can definitely merge, let me know... but looks like maybe we introduce a different issue in the arrow repo? I have not spotted this one before...

```
------------------------- ANTICONF ERROR ---------------------------
Configuration failed because arrow parquet was not found. Try installing:
 * brew: apache-arrow (Mac OSX)
If arrow parquet is already installed, check that 'pkg-config' is in your
PATH and PKG_CONFIG_PATH contains a arrow parquet.pc file. If pkg-config
is unavailable you can set INCLUDE_DIR and LIB_DIR manually via:
R CMD INSTALL --configure-vars='INCLUDE_DIR=... LIB_DIR=...'
```

I wonder why the Mac OSX anticonf is triggering.

nealrichardson:
It's not macOS; that error message appears only gives macOS help though. The actual problem is farther up. Arrow C++ is failing to build: https://travis-ci.org/rstudio/sparklyr/jobs/540843901#L2015

This is the same [issue](https://issues.apache.org/jira/browse/ARROW-5470) I [fixed](https://github.com/apache/arrow/pull/4443/files) for Arrow on Friday. Because your Travis builds are pinned to `dist: trusty`, you have an old version of Boost (1.54), which is no longer sufficient to build Arrow. 

To fix, we/you could either bump up to Xenial (the Travis default as of last month anyway) and follow the fix I did in https://github.com/apache/arrow/pull/4443/files, or try to find a way to install newer Boost on trusty (I'd guess there's a PPA for that), or I hear there is a vendored Boost lib in Arrow we could build with instead (though I haven't looked into what that entails). (See also https://issues.apache.org/jira/browse/ARROW-5474, the resolution of which I would expect to give advice on what to do.)

javierluraschi:
Oh that's great, that's for pointing out the issue, should be easy to update Travis, I can take a look.

PR #2032, opened 2019-06-05
Use xenial to fix deprecated boost version in arrow devel build

javierluraschi:


javierluraschi:
@nealrichardson seems `xenial` gets us closer but there seems to be some other exceptions against `arrow` devel, maybe caused by arrow maybe not. Here in spark-land we will have to investigate at some point, but sending your way in case you are already aware of this [cannot infer type from data ](https://github.com/apache/arrow/blob/d8bf1948ff19708ea2d6b77280252df51df545d3/r/src/array_from_vector.cpp#L800) error.

```
Context: copy data
sdf_copy_to works for default serializer: warning: `as_data_frame()` is deprecated, use `as_tibble()` (but mind the new semantics).
This warning is displayed once per session. 
sdf_copy_to works for default serializer: error: cannot infer type from data 
sdf_copy_to works for default serializer: 65.8194258213043
sdf_copy_to works for scala serializer: 3.12229371070862
sdf_copy_to works for csv serializer: 2.70926332473755
sdf_copy_to supports list of callbacks: error: cannot infer type from data 
spark_table_name() doesn't warn for multiline expression (#1386): 0.00432300567626953
sdf_copy_to supports list of callbacks: 0.276857137680054

Context: dbi
: error: cannot infer type from data 

Context: dplyr
: error: cannot infer type from data 

Context: ml clustering - kmeans
ml_kmeans() default params: 0.256178379058838

Context: describe
sdf_collect() works properly: error: cannot infer type from data 
sdf_collect() works properly: 0.276394605636597
sdf_collect() supports callback: 1.5267379283905
sdf_collect() supports callback expression: 0.87138843536377
```

nealrichardson:
Thanks @javierluraschi, traceback suggests that https://github.com/apache/arrow/pull/3565 (which landed Monday) may be implicated. @romainfrancois, can you take a look next week?

nealrichardson:
On a different note, I think you could make the `CHECK` warnings about long paths and invalid files go away if you went up a level (`..`) to clone and build arrow here: https://github.com/rstudio/sparklyr/blob/master/ci/arrow-build.sh#L3 so that you're not checking out arrow inside the sparklyr package dir.

PR #2039, opened 2019-06-10
Changes to add new qubole method to support sparklyr with Qubole cluster.

vipul1409:
Changes to add new qubole method to support sparklyr with Qubole cluster.

### DIFF
--- a/R/connection_spark.R
+++ b/R/connection_spark.R
@@ -221,7 +221,7 @@ spark_connect <- function(master,
 
   register_mapping_tables()
   
-  scon <- initialize_method(structure(method, class = "qubole"), scon)
+  scon <- initialize_method(structure(scon, class = method), scon)
 
   # notify connection viewer of connection
   libs <- c("sparklyr", extensions)
--- a/R/connection_spark.R
+++ b/R/connection_spark.R
@@ -241,6 +241,7 @@ spark_connect <- function(master,
   
   if (method == "qubole") {
     scon$config$sparklyr.web.spark <- invoke(scon$state$spark_context, "getSparkUIURL")
+    scon$config$sparklyr.web.yarn <- invoke(scon$state$spark_context, "getSparkUIURL")
   }
 
   # notify connection viewer of connection
--- a/R/connection_spark.R
+++ b/R/connection_spark.R
@@ -99,7 +99,7 @@ spark_config_shell_args <- function(config, master) {
 #' @export
 spark_connect <- function(master,
                           spark_home = Sys.getenv("SPARK_HOME"),
-                          method = c("shell", "livy", "databricks", "test"),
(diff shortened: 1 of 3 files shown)


javierluraschi:
Looks good; however, we need to scope all the qubole code in your own `qubole_connection.R` file. I'll add inline comments with the changes we need.

R/connection_spark.R
```diff
@@ -192,6 +195,24 @@ spark_connect <- function(master,
   } else if (method == "databricks") {
     scon <- databricks_connection(config = config,
                                   extensions)
+  } else if (method == "qubole") {
+    scon <- shell_connection(master = master,
```
javierluraschi: Instead of creating the `shell_connection` here, create a helper function `qubole_connection()` in a new `qubole_connection.R` file and move this code into it.

R/connection_spark.R
```diff
@@ -217,6 +238,13 @@ spark_connect <- function(master,
   # register mapping tables for spark.ml
 
   register_mapping_tables()
+  
+  if (method == "qubole") {
```
javierluraschi: Instead of special-casing here for `qubole`, create a new dispatcher function using S3 R methods. This is accomplished as follows:

1) In the bottom of this file add the main S3 dispatcher:

```r
initialize_method <- function(method, scon) {
   UseMethod("initialize_method")
}
```

2) Also add the default behavior in this file which to not do anything:

```r
initialize_method.default <- function(method, scon) {
   scon
}
```

3) Create a method object and dispatch the function by replacing the `if (method == "qubole")` with:

```r
scon <- initialize_method(structure(method, class = "qubole"), scon)
```

4) Implement your own initializer, in `qubole_connection.R` add this function:

```r
initialize_method.qubole <- function(method, scon) {
  scon$config$sparklyr.web.spark <- invoke(scon$state$spark_context,
                                             "getSparkUIURL")
  scon$config$sparklyr.web.yarn <- system("/usr/lib/zeppelin/bin/get_yarn_url.sh",
                                            intern = TRUE)
  scon
}
```

javierluraschi:
@vipul1409 two changes requested, if you need help with them, let me know to take a look.

R/connection_spark.R
javierluraschi: Actually... if you don't need changes at all in how the connection is initialized compared to the shell connection... you can also consider removing this code and changing instead:

```
if (method == "shell") {
```

into

```
if (method == "shell" || method == "qubole") {
```

javierluraschi:
@vipul1409 code changes looks great! Looks like Travis is complaining about the docs being out of sync. You would have to run "Tools" -> "Document" in RStudio or manually run:

```
devtools::document(roclets = c('rd', 'collate', 'namespace'))
```

javierluraschi:
Looking great, thanks!

PR #2048, opened 2019-06-24
Changed the path of get-yarn-url file in Qubole connector

AgrawalAmey:


R/qubole_connection.R
```diff
@@ -1,7 +1,7 @@
 initialize_method.qubole <- function(method, scon) {
   scon$config$sparklyr.web.spark <- invoke(scon$state$spark_context,
                                              "getSparkUIURL")
-  scon$config$sparklyr.web.yarn <- system("/usr/lib/zeppelin/bin/get_yarn_url.sh",
+  scon$config$sparklyr.web.yarn <- system("/usr/lib/rstudio-utils/bin/get-yarn-url.sh",
```
javierluraschi: You might want to consider something like:

```suggestion
  yarn_url <-  Sys.getenv("QUBOLE_YARN_WEB")
  scon$config$sparklyr.web.yarn <-if (nchar(yarn_url) > 0) yarn_url else system("/usr/lib/rstudio-utils/bin/get-yarn-url.sh",
```

That way, if this changes, you can simply override it in Qubole's cluster instead of having to rely on `sparklyr`.

javierluraschi:
Looks great, merging today.

AgrawalAmey:
Thank you.

PR #2118, opened 2019-08-23
Allow reading multiple files with `spark_read_`

jozefhajnala:
Some of the `org.apache.spark.sql.DataFrameReader` methods allow for passing `paths: String*`, besides `path: String`. We can use this to allow the `spark_read_` suite of functions to accept multiple paths, not just a single path and read multiple files in one call (inspired in part by [SO question](https://stackoverflow.com/q/57531647/8876652)).

This PR proposes to add this support to:

- `spark_read_parquet()`
- `spark_read_json()`
- `spark_read_text()` - with `whole=FALSE` only, stops with a meaningful error message if multiple paths are provided with `whole=TRUE`
- `spark_read_orc()`

Unit tests are also added for the above reading functions.

Compatibility notes on [DataFrameReader](https://spark.apache.org/docs/2.3.0/api/scala/index.html#org.apache.spark.sql.DataFrameReader) supporting `paths: String*`:

- `.parquet` - since 1.4.0 
- `.json` - since 2.0.0
- `.text` - since 1.6.0
- `.orc` - since 2.0.0

### DIFF
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -244,7 +244,7 @@ spark_read_parquet <- function(sc,
   path <- params[-1L]
   if (overwrite) spark_remove_table_if_exists(sc, name)
 
-  df <- spark_data_read_generic(sc, lapply(path, spark_normalize_path), "parquet", options, columns, schema)
+  df <- spark_data_read_generic(sc, as.list(spark_normalize_path(path)), "parquet", options, columns, schema)
   spark_partition_register_df(sc, df, name, repartition, memory)
 }
 
@@ -328,7 +328,7 @@ spark_read_json <- function(sc,
   path <- params[-1L]
   if (overwrite) spark_remove_table_if_exists(sc, name)
 
-  df <- spark_data_read_generic(sc, lapply(path, spark_normalize_path), "json", options, columns)
+  df <- spark_data_read_generic(sc, as.list(spark_normalize_path(path)), "json", options, columns)
   spark_partition_register_df(sc, df, name, repartition, memory)
 }
 
@@ -803,7 +803,7 @@ spark_read_text <- function(sc,
     df <- invoke(hive_context(sc), "createDataFrame", rdd, schema)
   }
   else {
-    df <- spark_data_read_generic(sc, lapply(path, spark_normalize_path), "text", options, columns)
+    df <- spark_data_read_generic(sc, as.list(spark_normalize_path(path)), "text", options, columns)
   }
 
   spark_partition_register_df(sc, df, name, repartition, memory)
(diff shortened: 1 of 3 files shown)


R/data_interface.R
```diff
@@ -239,10 +239,12 @@ spark_read_parquet <- function(sc,
                                columns = NULL,
                                schema = NULL,
                                ...) {
-  c(name, path) %<-% spark_read_compat_param(sc, name, path)
+  params <- spark_read_compat_param(sc, name, path)
+  name <- params[1L]
+  path <- params[-1L]
```
javierluraschi: Did you hit a particular issue with:

```r
c(name, path) %<-% spark_read_compat_param(sc, name, path)
```

Otherwise, can we revert this and other changes to `c(name, path) %<-%`?

R/data_interface.R
```diff
@@ -239,10 +239,12 @@ spark_read_parquet <- function(sc,
...
   if (overwrite) spark_remove_table_if_exists(sc, name)
 
-  df <- spark_data_read_generic(sc, list(spark_normalize_path(path)), "parquet", options, columns, schema)
+  df <- spark_data_read_generic(sc, lapply(path, spark_normalize_path), "parquet", options, columns, schema)
```
javierluraschi: It would also make sense to vectorize `spark_normalize_path` to support vectors, I would change this: https://github.com/rstudio/sparklyr/blob/787fe0b53dd402490fde0f32ac48f4b4d1af1861/R/utils.R#L151-L159 to:

```r
spark_normalize_single_path <- function(path) {
  # don't normalize paths that are urls
  if (grepl("[a-zA-Z]+://", path)) {
    path
  }
  else {
    normalizePath(path, mustWork = FALSE)
  }
}

spark_normalize_path <- function(paths) {
  unname(sapply(paths, spark_normalize_single_path))
}
```

R/data_interface.R
jozefhajnala: Yes, there is a check in `%<-%` that would fail here as it would expect length 2 on both sides:
```
if (length(dest) != length(value)) stop("Assignment must contain same number of elements")
```

Interestingly enough, the [zeallot package](https://github.com/r-lib/zeallot) implements the same operator and has a construct for similar cases:
```
c(first, ...rest) %<-% list("a", "b", "c", "d", "e")
```
It also is a (recursive) dependency of sparklyr, but I did not want to make such a strong change here, so I went with the base solution above.

javierluraschi:
@jozefhajnala this is pretty great! Thank you! Left a few comments, if you are low on time let us know and we can iterate over your Pr.

kevinykuo:
@jozefhajnala do you mind adding tests for reading multiple files with globbing expressions to make sure we don't break those?

R/data_interface.R
jozefhajnala: Thanks for the comment, I give this a shot in https://github.com/rstudio/sparklyr/pull/2118/commits/d075bd48313511bfe39fa2f210e688e9837fb30a, looks fine to me, at least all tests pass.

jozefhajnala:
> @jozefhajnala do you mind adding tests for reading multiple files with globbing expressions to make sure we don't break those?

Added some in https://github.com/rstudio/sparklyr/pull/2118/commits/7b0cccef7c63382aae7a11874e36cf39a8de2826.

javierluraschi:
Thanks @jozefhajnala!

PR #2132, opened 2019-09-20
Fix Arrow devel tests

nealrichardson:
This attempts to fix the failing Arrow devel build, which is caused by the moving of the record batch stream reader function to be a method on the R6 class. 

The build is still failing for me, though not as hard as on rstudio/master. It's hard for me to read the traceback though, so maybe you can see what's missing.

I also changed the travis config so that arrow devel is no longer an allowed failure. We shouldn't be breaking things and we should learn about it as soon as we do break something.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,19 +92,6 @@ matrix:
         - r-lib/ellipsis
   allow_failures:
     - env: R_DEVEL_PACKAGES="true"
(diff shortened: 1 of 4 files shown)


javierluraschi:
We can merge with https://github.com/rstudio/sparklyr/pull/2133, thank you!

Issue #2104, opened 2019-07-31
sparklyr integration with Hive on Hadoop 3


suraggupta:
How do we use `sparklyr` to query tables from Hive in Hadoop 3, now that catalogs for Apache Hive and Apache Spark are separated, and Spark SQL no longer seems to work? Is this something that's being worked on?

javierluraschi:
This is not under our radar, could you share some specifics of your environment? Which Spark distribution are you using? Version of Spark and Hadoop? etc. It would be ideal if you could provide for you us to reproduce this... that said, my guess is that Spark needs to be properly configureed to use the Hive catalog, I'm not sure this is specific to `sparklyr`, is it?

For instance, have you tried retrieving data from the Hive catalog using `pyspark`? Does this work? My guess is that this wouldn't work either, which would mean the Spark configuration needs to change, if it works in Python but not in R, let us know to take a look!

SteGerm:
Hi,

Since this topic is still opened and the (broad) topic match my issue I'll add some specifics about an issue we encounter on our cluster that was juste upgraded (up to HDP 3.1.4 / Hive 3.1 / Spark 2.3.2 / SparklyR 1.0.2 / YARN).
Using the following configurations (same file used to configure pyspark sessions which are working btw) we managed to have access to Hive Metastore and as such can list DB, Tables, Tables structures, BUT we can't access to data for every new managed tables. (External tables works just fine).
ie. Querying a managed table result in success query displaying expected columns but without any values.
(same query works just fine in pyspark/spark-shell and every other mean)

Configurations/Connexion (Contains some dynamic allocation configuration but it is NOT enabled on this test):
> # Import
> library(sparklyr)
> library(dplyr)
> library(DBI)
> 
> # Connexion Spark
> Sys.setenv(SPARK_HOME= "/usr/hdp/current/spark2-client")
> Sys.setenv(YARN_CONF_DIR = "/etc/hadoop/conf")
> Sys.setenv(HADOOP_CONF_DIR = "/etc/hive/conf")
> Sys.setenv(HIVE_CONF_DIR = "/etc/hive/conf")
> .libPaths( c(file.path(Sys.getenv("SPARK_HOME"), "R", "lib"), .libPaths()   ))
> 
> conf <- config::get(file = "/app/spark/conf/config_init_spark_latest.yaml", 
+                     config = "small_light")
> 
> conf$sparklyr.shell.files <- '/etc/hive/conf/hive-site.xml'
> conf
[...]

PR #2146, opened 2019-10-16
Some cloudera clusters have been failing with this override

javierluraschi:
Fixes:

```
Error: org.apache.spark.sql.AnalysisException: java.lang.RuntimeException: java.io.IOException: Permission denied;
        at org.apache.spark.sql.hive.HiveExternalCatalog.withClient(HiveExternalCatalog.scala:108)
        at org.apache.spark.sql.hive.HiveExternalCatalog.databaseExists(HiveExternalCatalog.scala:216)
        at org.apache.spark.sql.internal.SharedState.externalCatalog$lzycompute(SharedState.scala:114)
        at org.apache.spark.sql.internal.SharedState.externalCatalog(SharedState.scala:102)
        at org.apache.spark.sql.internal.SharedState.globalTempViewManager$lzycompute(SharedState.scala:141)
        at org.apache.spark.sql.internal.SharedState.globalTempViewManager(SharedState.scala:136)
        at org.apache.spark.sql.hive.HiveSessionStateBuilder$$anonfun$2.apply(HiveSessionStateBuilder.scala:55)
        at org.apache.spark.sql.hive.HiveSessionStateBuilder$$anonfun$2.apply(HiveSessionStateBuilder.scala:55)
        at org.apache.spark.sql.catalyst.catalog.SessionCatalog.g
```

CC: @kevinykuo

### DIFF
--- a/inst/conf/config-template.yml
+++ b/inst/conf/config-template.yml
@@ -5,9 +5,6 @@ default:
   # include the embedded csv package for spark 1.x
   sparklyr.connect.csv.embedded: "^1.*"
 
-  # Enable Hive support by default.
-  spark.sql.catalogImplementation: "hive"
-
   # default spark packages to load
   # sparklyr.defaultPackages:
 


kevinykuo:
This change was needed as a fix for https://github.com/rstudio/sparklyr/issues/1496. We know it was working for folks on Hadoop 2.x (CDH < 6.0, HDP <3.0) and I'm concerned that reverting this outright might break some clusters. I think if the issue is reproducible on Hadoop 3 and the PR fixes it (may be related to https://github.com/rstudio/sparklyr/issues/2104), we can add a switch on the hadoop version to be safe.

cc @edgararuiz

javierluraschi:
I agree! Although, we want to to check that's the case by merging in `devel` and hopefully getting some community feedback if this change backfires. I think in general we don't want to override Spark configuration settings, but if we do, it would be nice to better understand exactly why us (but not other interfaces) require this override.

AgrawalAmey:
Just an FYI, as @kevinykuo suggested this change broke sparklyr on Qubole clusters (which run one Hadoop 2.x). We are adding the override on our end to handle this.

alexvorobiev:
This broke sparklyr on MapR cluster.

javierluraschi:
@alexvorobiev there is a workaround in https://github.com/sparklyr/sparklyr/issues/2202 as in, to add:

```
config$spark.sql.catalogImplementation <- "hive"
```

alexvorobiev:
@javierluraschi Thanks! Is it documented anywhere? I guess anybody who is on MapR has to use that.

Issue #2152, opened 2019-10-22
Allow reading config file path from environment variable


AgrawalAmey:
Right now the only way to set config file is to pass the function parameter, hence,  user has to specify the location of the file each time while running spark_connect. Reading the config path file to be read from an environment variable would allow the user to specify it once in .bashrc and use it out of the box.

PR #2153, opened 2019-10-22
Added SPARKLYR_CONFIG_FILE_PATH environment variable as an option to reading sparklyr config

AgrawalAmey:
Fix for issue #2152

### DIFF
--- a/R/config_spark.R
+++ b/R/config_spark.R
@@ -22,7 +22,7 @@ spark_config <- function(file = "config.yml", use_default = TRUE) {
   optionsConfig <- options()[optionsConfigCheck]
   baseConfig <- merge_lists(optionsConfig, baseConfig)
 
-  userEnvConfig <- tryCatch(config::get(file = Sys.getenv("SPARKLYR_CONFIG_FILE_PATH")), error = function(e) NULL)
+  userEnvConfig <- tryCatch(config::get(file = Sys.getenv("SPARKLYR_CONFIG_FILE")), error = function(e) NULL)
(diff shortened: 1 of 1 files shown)


javierluraschi:
Notice that there is also `R_CONFIG_FILE` which is defined in the `config` package and should have a similar behavior: https://github.com/rstudio/config/blob/master/R/get.R#L27

But it also makes sense to have an environment variable for `sparklyr`, would suggest we name this `SPARKLYR_CONFIG_FILE`.

R/config_spark.R
```diff
@@ -22,9 +22,12 @@ spark_config <- function(file = "config.yml", use_default = TRUE) {
   optionsConfig <- options()[optionsConfigCheck]
   baseConfig <- merge_lists(optionsConfig, baseConfig)
 
+  userEnvConfig <- tryCatch(config::get(file = Sys.getenv("SPARKLYR_CONFIG_FILE_PATH")), error = function(e) NULL)
```
javierluraschi:
```suggestion
  userEnvConfig <- tryCatch(config::get(file = Sys.getenv("SPARKLYR_CONFIG_FILE")), error = function(e) NULL)
```

javierluraschi:
Notice that there is also `R_CONFIG_FILE` which is defined in the `config` package and should have a similar behavior: https://github.com/rstudio/config/blob/master/R/get.R#L27

But it also makes sense to have an environment variable for `sparklyr`, would suggest we name this `SPARKLYR_CONFIG_FILE`.

javierluraschi:
Thank you, @AgrawalAmey!

PR #2154, opened 2019-10-22
Support spark 2.4 with scala-2.12 

lu-wang-dl:
Modify `addTaskCompletionListener` so that we can generate the sparklyr jar files with spark 2.4 and support to compile with both scala 2.11 and scala 2.12.


javierluraschi:
LGTM. Thanks @lu-wang-dl! I'll add the change to the `NEWS` file in a follow up commit but would help to include that in future PRs.

PR #2155, opened 2019-10-23
Send R environments to Scala maps

### LINKED ISSUE #1058: Conversion between R and Scala
Dear sparklyr team,

according to some tries and errors on my side, I found out that 
- a ``scala Map`` seems to be well translateable to a ``R named list``
- however a ``R named list`` will be translated to an java Array which is recognized as ``scala Seq`` of respective Type, i.e. it looses its name information
- a ``R environment`` in fact gets translated to ``java Map``, which however unfortunately does not translate that easily to a scala Map

These three observations together suggest to me that:
1. ``scala Maps`` as well ``java Maps`` should be translated to ``R environments`` instead of ``R lists``
1. ``scala Maps`` might be very much preferable over ``java Maps`` as the translation of ``R environments``, just because all suggested examples of extending sparklyr use scala code and not java

what do you think?

kevinykuo:
In this change:

- R environments are now sent to Scala Maps rather than `java.util.Map[Object, Object]`. This is a breaking change for users that depended on the old behavior, but should be minimal if not nonexistent. Closes #1058 which unblocks a few use cases.


kevinykuo:
@dkincaid could you confirm this change resolves the issue with `nlp_text_matcher()`?

dkincaid:
Looks like it's working. I installed sparklyr from kevinykuo/sparklyr@feature/map-serialize and tested using the following (as well as my text_matcher tests):

```r
library(sparklyr)
library(sparknlp)
sc <- spark_connect(master = "local", version = "2.4.0")
jobj <- invoke_new(sc, "com.johnsnowlabs.nlp.annotators.TextMatcher")
opts <- list2env(list(format = "text"))
invoke(jobj, "setEntities", "/tmp/path", sparknlp:::read_as("LINE_BY_LINE"), opts)
```

javierluraschi:
Another workaround would be to create a wrapper function in `sparknlp` that converts a Java Map into a Scala Map; but regardless, I think it makes more sense to default to Scala Maps, not Java Maps, so this PR LGTM.

I would remove the `(breaking change)` statement, users shouldn't care and we will have to make sure extensions are not broken next time we release. If this change backfires while testing extensions, we would have to revert this change and add a new helper Scala function in `sparknlp`, but I think this is unlikely.

kevinykuo:
@javierluraschi could you elaborate on
> Another workaround would be to create a wrapper function in sparknlp that converts a Java Map into a Scala Map

Do you mean that every Scala function that takes a map would need its own custom wrapper function? Or is there a simpler way we're missing here?

javierluraschi:
Yeah, we would need a `sparkNlpSetEntities()` that converts to Scala Maps before making the call to setEntities(). Not sure how many calls are in this situation for SparkNLP, but yeah, we would need one for each call that requires a Scala Map parameter.

I think this PR is totally fine, defaulting to Scala maps makes sense, if it doesn't break anything.

Another option would be to add support for a `convert = FALSE` parameter in `invoke()` which would not convert back to R types but rather retrieve a `jobj`. I think this is needed also in other packages like `reticualte` since sometimes the conversion code gets tricky so it's better to handle it in R. We would need to add a parameter here:

https://github.com/rstudio/sparklyr/blob/fe567e450b4716c8ba3a7beaea5c5103151df503/inst/livy/spark-1.5.2/handler.scala#L29-L32

And then force the serializer to write back a `jobj` in here: https://github.com/rstudio/sparklyr/blob/68fe0bc1f235ce3bc12ac10cf08cb5cbc3e7b95e/inst/livy/spark-1.5.2/serializer.scala#L230-L240

I've opened to consider supporting this: https://github.com/rstudio/sparklyr/issues/2156

kevinykuo:
Made the recommended change to `NEWS` entry.

Merging! (Let me know if this ends up breaking the book)

PR #2163, opened 2019-11-09
[WIP] Specify dplyr package for `filter(...)`

Loquats:
Specify dplyr package for `filter(...)`

### DIFF
--- a/tests/testthat/test-broom-linear_svc.R
+++ b/tests/testthat/test-broom-linear_svc.R
@@ -24,7 +24,7 @@ test_that("linear_svc.augment() works", {
   iris_tbl <- testthat_tbl("iris")
 
   iris_tbl2 <- iris_tbl %>%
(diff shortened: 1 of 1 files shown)


Loquats:
Actually, before I made this change, this test passes when I run it on command line using
```
$ R
> library(devtools)
> test_file("tests/testthat/test-broom-linear_svc.R")
```
But it does not pass when I run `test_file("tests/testthat/test-broom-linear_svc.R")` on the RStudio repl. If I do not make this change, the test fails with error `object 'Species' not found`. Perhaps there is something misconfigured on my RStudio.

javierluraschi:
Other tests run [test_requires("dplyr")](https://github.com/rstudio/sparklyr/blob/0be38f0205ff61089dfbbdb6098078ef726365d1/tests/testthat/test-dplyr.R#L5), either way it's fine. Thanks for the PR!

PR #2176, opened 2019-11-18
support sparklyr executor memory setting and warn to fix printed book issue

javierluraschi:
The printed version of Spark with R book suggests connecting in YARN as:

```
# Memory in Driver
config["sparklyr.shell.driver-memory"] <- "2g"

# Memory per Worker
config["sparklyr.executor.memory"] <- "2G" 

# Cores per Worker
config["sparklyr.shell.executor-cores"] <- 1

# Number of Workers
config["sparklyr.shell.num-executors"] <- 3
```

which is honestly kinda intuitive since one is specifying all settings with `sparklyr` prefix; however, this is incorrect. To help out users, we will support `sparklyr.shell.executor-cores`, but warn users this is the incorrect setting.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
-Version: 1.0.5
+Version: 1.0.5.9000
 Authors@R: c(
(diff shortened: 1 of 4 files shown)


javierluraschi:
Actually, not really... my bad, closing.

Issue #2172, opened 2019-11-17
API changes for OneHotEncoder and OneHotEncoderEstimator in spark 3.0


lu-wang-dl:
In spark 3.0, `OneHotEncoder` which is deprecated in 2.3, is removed and `OneHotEncoderEstimator` is now renamed to `OneHotEncoder`.

PR #2178, opened 2019-11-18
Update ft_one_hot_encoder to be compatible with spark 3.0

lu-wang-dl:
This is to fix the #2172 with spark 3.0.
- Update `ft_one_hot_encoder`with spark 3.0 to match the new API.
- Dreprecate `ft_one_hot_encoder_estimator` with spark 3.0 by updating `spark_require_version` with `required_max`
- Skip the tests with spark 3.0 in `test-feature-one-hot-encoder-estimator.R`
- Update the tests in `test-ml-feature-one-hot-encoder.R` only check the arg in spark 3.0 or later.

### DIFF
--- a/R/ml_feature_one_hot_encoder.R
+++ b/R/ml_feature_one_hot_encoder.R
@@ -33,6 +33,7 @@ ft_one_hot_encoder.spark_connection <- function(x, input_col = NULL, output_col 
     drop_last = drop_last,
     uid = uid
   ) %>%
+    c(rlang::dots_list(...)) %>%
     validator_ml_one_hot_encoder()
   version <- spark_version(x)
   if(version < "3.0.0") {
@@ -124,8 +125,6 @@ ft_one_hot_encoder.tbl_spark <- function(x, input_col = NULL, output_col = NULL,
   } else {
     ml_fit_and_transform(stage, x)
   }
-
-  # ml_transform(stage, x)
 }
 
 new_ml_one_hot_encoder <- function(jobj) {
--- a/R/ml_feature_one_hot_encoder.R
+++ b/R/ml_feature_one_hot_encoder.R
@@ -11,8 +11,7 @@
 #' @param drop_last Whether to drop the last category. Defaults to \code{TRUE}.
 #'
 #' @export
-ft_one_hot_encoder <- function(x, input_col = NULL, output_col = NULL,
-                               input_cols = NULL, output_cols = NULL, handle_invalid = NULL,
+ft_one_hot_encoder <- function(x, input_cols = NULL, output_cols = NULL, handle_invalid = NULL,
                                drop_last = TRUE, uid = random_string("one_hot_encoder_"), ...) {
   check_dots_used()
   UseMethod("ft_one_hot_encoder")
@@ -21,12 +20,9 @@ ft_one_hot_encoder <- function(x, input_col = NULL, output_col = NULL,
 ml_one_hot_encoder <- ft_one_hot_encoder
 
 #' @export
-ft_one_hot_encoder.spark_connection <- function(x, input_col = NULL, output_col = NULL,
-                                                input_cols = NULL, output_cols = NULL, handle_invalid = "error",
+ft_one_hot_encoder.spark_connection <- function(x, input_cols = NULL, output_cols = NULL, handle_invalid = "error",
                                                 drop_last = TRUE, uid = random_string("one_hot_encoder_"), ...) {
   .args <- list(
-    input_col = input_col,
-    output_col = output_col,
     input_cols = input_cols,
     output_cols = output_cols,
     handle_invalid = handle_invalid,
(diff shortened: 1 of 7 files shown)


tests/testthat/test-ml-feature-one-hot-encoder.R
```diff
@@ -1,33 +1,32 @@
...
 })
 
 test_that("ft_one_hot_encoder() works", {
-  skip_on_spark_master()
   sc <- testthat_spark_connection()
   iris_tbl <- testthat_tbl("iris")
   expect_equal(
     iris_tbl %>%
-      ft_string_indexer("Species", "indexed") %>%
+      ft_string_indexer("Species", "indexed", string_order_type = "alphabetDesc") %>%
```
lu-wang-dl: Modify this because of the behavior changes for StringIndexer. 
Check: https://github.com/apache/spark/blob/7d4eb38bbcc887fb61ba7344df3f77a046ad77f8/docs/ml-migration-guide.md#changes-of-behavior

kevinykuo:
Can we just keep the plural, i.e., `input_cols`/`output_cols`, params?

lu-wang-dl:
Since the new `OneHotEncoder` has single-column input/ouput support (# SPARK-29565). DO we want to keep the API consistent with spark? 

https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/feature/OneHotEncoder.scala

kevinykuo:
We definitely want to keep interop in mind, especially since we support reading/loading pipelines. However, I wonder if this is a case where we can get away with simplifying the API a bit. Is there anything preventing us from taking a single column name in `input_cols` for example?

lu-wang-dl:
For the new oneHotEncoder, I think we can use inputCols and outputCols only. But to be compatible with spark 2.x, we still need to keep inputCol and outputCol. So we didn't really simplify the API here.

kevinykuo:
For spark 2.x we can just check that a single column name is provided and invoke the appropriate setter e.g. `setInputCol`, no?

R/ml_feature_one_hot_encoder.R
```diff
@@ -21,59 +22,130 @@ ml_one_hot_encoder <- ft_one_hot_encoder
...
                                            drop_last = TRUE, uid = random_string("one_hot_encoder_"), ...) {
-  stage <- ft_one_hot_encoder.spark_connection(
-    x = spark_connection(x),
-    input_col = input_col,
-    output_col = output_col,
-    drop_last = drop_last,
-    uid = uid,
-    ...
-  )
+  sc <- spark_connection(x)
```
falaki: These three seem to be a useful combination. How about adding a new utility function like this:
```
is_spark_3 <- function() {
  version <- spark_connection() %>% version 
  return version >= "3"
}
```

R/ml_feature_one_hot_encoder.R
```diff
@@ -21,59 +22,130 @@ ml_one_hot_encoder <- ft_one_hot_encoder
...
+  version <- spark_version(sc)
+  version <- "3.0.0"
```
falaki: Is this for debugging?

lu-wang-dl:
@kevinykuo Do you mean leave `inputCol` and `outputCol` as the extra params? Would this change the user behavior?
What happens if the user calls `ft_one_hot_encoder("indexed", "encoded")` ?

kevinykuo:
@lu-wang-dl I mean we'd update the signature to be `ft_one_hot_encoder <- function(x, input_cols = NULL, output_cols = NULL, ...)`. If the user calls `ft_one_hot_encoder(., "indexed", "encoded")` in Spark 2.x it'd still work.

lu-wang-dl:
But in spark 2.x, there is no params `input_cols` and `output_cols` in one_hot_encoder, it calls 
```
spark_pipeline_stage(
      x, "org.apache.spark.ml.feature.OneHotEncoder",
      input_col = .args[["input_col"]], output_col = .args[["output_col"]], uid = .args[["uid"]]
```
Do I miss understand something?

kevinykuo:
If you do `input_col = .args[["input_cols"]], output_col = .args[["output_cols"]]` it'd probably work. Would be good to check the length and yell at the user if they try to pass multiple columns in spark 2.x.

lu-wang-dl:
Then we probably should warn the user that they cannot call 
`ft_one_hot_encoder(., input_col = "indexed", output_col = "encoded")`.

kevinykuo:
`ft_one_hot_encoder(., input_col = "indexed", output_col = "encoded")` seems pretty harmless?

kevinykuo:
(Note that R function argument names can be partially matched.)

lu-wang-dl:
Thanks. I will update the PR.

javierluraschi:
(( For others following along the `OneHotEncoderEstimator` deprecation, see [Java docs 3.0 preview](https://spark.apache.org/docs/3.0.0-preview/api/java/index.html) and [Spark 3.0 preview release feature list and major changes](http://apache-spark-developers-list.1001551.n3.nabble.com/Spark-3-0-preview-release-feature-list-and-major-changes-td28050.html). ))

R/ml_feature_one_hot_encoder.R
```diff
@@ -20,60 +20,119 @@ ft_one_hot_encoder <- function(x, input_col = NULL, output_col = NULL,
...
+    estimator <- spark_pipeline_stage(
+      x, "org.apache.spark.ml.feature.OneHotEncoder",
+      input_cols = .args[["input_cols"]], output_cols = .args[["output_cols"]], uid = .args[["uid"]]
+    ) %>%
+      invoke("setHandleInvalid", .args[["handle_invalid"]]) %>%
+      invoke("setDropLast", .args[["drop_last"]]) %>%
+      new_ml_one_hot_encoder()
+  } else {
+    if (lengths(.args[["input_cols"]]) > 1 || lengths(.args[["output_cols"]]) > 1) {
+      stop("OneHotEncoder does not support encoding multiple columns", call. = FALSE)
```
kevinykuo: Can we add a test for this error message?

R/ml_feature_one_hot_encoder.R
```diff
@@ -20,60 +20,119 @@ ft_one_hot_encoder <- function(x, input_col = NULL, output_col = NULL,
...
+    ml_fit_and_transform(stage, x)
+  }
 }
 
 new_ml_one_hot_encoder <- function(jobj) {
   new_ml_transformer(jobj, class = "ml_one_hot_encoder")
 }
 
+new_ml_one_hot_encoder_model <- function(jobj) {
+  spark_require_version(x, "3.0.0")
```
javierluraschi: Notice that `x` here is undefined, replace `x` with `spark_connection(jobj)`. Fixing this should also fix the last remaining test.

javierluraschi:
From the tests,

```
new_ml_one_hot_encoder_model: no visible binding for global variable
  ‘x’
Undefined global functions or variables:
  x
```

Which this comment should fix: https://github.com/rstudio/sparklyr/pull/2178/files#r348239512

Issue #2181, opened 2019-11-19
New error handling logic in ml.feature.QuantileDiscretizer in spark 3.0


lu-wang-dl:
In spark 3.0, it updates multi-column error handling logic in QuantileDiscretizer
https://github.com/apache/spark/blob/bb478706b54cb247fe541ce0f860a7874ffbd42f/mllib/src/main/scala/org/apache/spark/ml/feature/QuantileDiscretizer.scala#L177

which causes the failure in https://github.com/rstudio/sparklyr/blob/master/tests/testthat/test-ml-feature-quantile-discretizer.R#L47

```r
test-ml-feature-quantile-discretizer.R:56: error: ft_quantile_discretizer works on multiple columns

java.lang.IllegalArgumentException: requirement failed: 
exactly one of numBuckets, numBucketsArray Params to be set, but both are set.
```

PR #2191, opened 2019-11-19
Change the way to set NumBuckets in QuantileDiscretizer

lu-wang-dl:
This PR is to fix the issue #2181  .


javierluraschi:
This PR looks ready, I usually restart the jobs that fail when it looks unrelated. I'll merge this one later today after the test succeed.

PR #2203, opened 2019-11-22
Remove compute_cost in spark 3.0

### LINKED ISSUE #2180: computeCost is removed from org.apache.spark.ml.clustering.KMeansModel in Spark 3.0
`computeCost` is deprecated and removed in 3.0.0. It causes the failure in https://github.com/rstudio/sparklyr/blob/master/tests/testthat/test-ml-clustering-kmeans-ext.R#L87

lu-wang-dl:
This is to fix #2180 .

### DIFF
--- a/R/ml_clustering_kmeans.R
+++ b/R/ml_clustering_kmeans.R
@@ -132,9 +132,6 @@ new_ml_kmeans <- function(jobj) {
 }
 
 new_ml_kmeans_model <- function(jobj) {
-  sc <- spark_connection(jobj)
-  version <- spark_version(sc)
-
   summary <- possibly_null(~ new_ml_kmeans_summary(invoke(jobj, "summary")))()
   kmeans_model <- new_ml_clustering_model(
     jobj,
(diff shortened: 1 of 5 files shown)


R/ml_clustering_kmeans.R
```diff
@@ -133,18 +133,23 @@ new_ml_kmeans <- function(jobj) {
...
       ~ invoke(jobj, "clusterCenters") %>%
         purrr::map(invoke, "toArray")
     ),
-    compute_cost = function(dataset) {
-      invoke(jobj, "computeCost", spark_dataframe(dataset))
-    },
     summary = summary,
     class = "ml_kmeans_model")
+
+  if (!is_required_spark(jobj, "3.0.0")) {
```
javierluraschi: I think we would want to warn here for Spark >= 3.0.0, something like:

```
warning("The 'compute_cost' parameter is no longer supported in Spark 3.x")
```

If we think there might be a few more parameter deprecations, worth calling and adding:

```
spark_param_deprecated <- function(param, version = "3.x") {
  warning("The '", param, "' parameter is deprecated in Spark ", version)
}
```

PR #2213, opened 2019-12-06
Implement connection method for Databricks Connect

Loquats:


tests/testthat/test-databricks-connect.R
```diff
@@ -0,0 +1,23 @@
+context("databricks-connect")
+
+Sys.setenv(TEST_DATABRICKS_CONNECT = 'true')
```
lu-wang-dl: We should only run this test for databricks-connect instead of all different settings in sparklyr, right? Currenly the travis test fails on all different setups.

Loquats: I tried to make it run even on travis, and just use a fake databricks-connect connection, but it does not work. So I have changed it to only run on Jenkins when we are testing dbconnect.

R/connection_spark.R
```diff
@@ -87,6 +87,10 @@ spark_config_shell_args <- function(config, master) {
   }))
 }
 
+no_databricks_guid <- function() {
+  mget("DATABRICKS_GUID", envir = .GlobalEnv, ifnotfound="") == ""
```
falaki: Nit: `ifnotfound = ""`

tests/testthat/helper-initialize.R
```diff
@@ -338,3 +342,8 @@ skip_on_spark_master <- function() {
   is_on_master <- identical(Sys.getenv("SPARK_VERSION"), "master")
   if (is_on_master) skip("Test skipped on spark master")
 }
+
+skip_unless_databricks_connect <- function() {
+  if (Sys.getenv("TEST_DATABRICKS_CONNECT") != "true")
+    skip("Test only run on Databricks Connect")
```
falaki: Nit: `Test only runs`

R/data_copy.R
```diff
@@ -205,7 +205,7 @@ spark_data_copy <- function(
     stop("The repartition parameter must be an integer")
   }
 
-  if (!spark_connection_is_local(sc) && identical(serializer, "csv_file")) {
+  if ((!spark_connection_is_local(sc) || sc$method == "databricks-connect") && identical(serializer, "csv_file")) {
```
falaki: Is it safer to update `spark_conection_is_local()` to return false when we are running on Databricks Connect? I wonder what other cases have different behavior. Would be good to check.

falaki:
LGTM waiting for CI before I merge.

PR #2215, opened 2019-12-06
Turn arrow cmake flags on

nealrichardson:
Recent changes in `apache/arrow`, in order to make the default build more minimal and thus easier to build, meant that certain features that were `ON` by default (or even not able to be toggled on/off) are now `OFF` unless you enable them. (See e.g. https://github.com/apache/arrow/commit/d6caca311356d557507b8d03526655f8184fa16c)

This PR should fix arrow dev failures on sparklyr Travis, such as https://travis-ci.org/sparklyr/sparklyr/jobs/621452801.


javierluraschi:
Thank @nealrichardson! We just enabled https://developercertificate.org/ in GitHub which, from now on, will require us to perform commits as `git commit -s -m "<message>"` (`-s` stamps them with this certificate of origin).

PR #2216, opened 2019-12-08
Include barrier in spark_apply

samuelmacedo83:
I added barrier to spark_apply, but there are some points to discuss:

- Pass `columns` parameter to spark_apply ( error in `schema <- spark_schema_from_rdd(sc, rdd, columns)`)
- `sdf_register` gives an error when use barriers, that is why I use `invoke("collect") in the end
- `worker_log_env$sessionId` is `NULL`, that is why `spark_log(sc, filter = "IPs:")` returns nothing

Here is an example you can run
```
sc <- spark_connect(master = "local[3]")
sdf_len(sc, 3, repartition = 3) %>% 
  spark_apply(nrow, barrier = TRUE, columns = c(id = "integer")) %>% 
  invoke("collect")

```


javierluraschi:
Just FYI, I'm working on a few suggestions here, will send them out in a bit.

javierluraschi:
Push a couple commits to move the `address` / barrier context to the closure to enable:

```r
library(sparklyr)

sdf_len(sc, 1, repartition = 1) %>%
  spark_apply(~ .y, barrier = TRUE, columns = c(address = "character")) %>%
  collect()
```
```
# A tibble: 1 x 1
  address        
  <chr>          
1 localhost:50693
```

javierluraschi:
To rebuild the jars we will need to merge https://github.com/sparklyr/sparklyr/pull/2218 since Spark 1.X build is broken... In the meantime, this PR could use some tests for the example I mentioned before, probably also entry in the `NEWS`, I would add:

```
### Distributed R

- Add support for barrier execution mode with `barrier = TRUE` in `spark_apply()` (@samuelmacedo83, #2216).
```

PR #2217, opened 2019-12-09
Switch Livy tests to use jars

javierluraschi:
Currently, we run tests using source-code-based Livy (which uploads the sparklyr backend as a source); however, we want to deprecate this mode in Spark 3.0 to simplify our codebase. This PR specifies the version with `method = "livy"` which would make the connection use the correct GitHub jar when connecting.


javierluraschi:
Need to fix:

```
Livy installed.: error: invalid version specification ‘master’ 
```

PR #2219, opened 2019-12-10
Improve documentation for the `columns` argument

jozefhajnala:
This proposes a small improvement to the docs of the `read_` functions in case the user wants to specify the `columns` argument. It provides an example mapping for each of `spark.sql.types`.


javierluraschi:
Thanks for approving, @vipul1409 !

PR #2234, opened 2020-01-06
Deprecate Livy source-uploads to favor using JARs

javierluraschi:
Earlier versions of sparklyr supported connecting to Livy as:

```r
sc <- spark_connect(
  master = "https://domain/livy/",
  method = "livy",
  config = livy_config(user = "<user>", password = "<password>"))
```

Since the Spark version was unknown, this would cause `sparklyr` to connect, check the Spark version, and then upload all correct sources to Livy (see [inst/livy/](https://github.com/sparklyr/sparklyr/tree/feature/sparklyr-1.0/inst/livy/)), which at the time was useful to ensure connectivity to any Livy instance.

However, in order to support this source-based upload to Livy, the `sparklyr` backend code had to be written to be serializable. This was OK in early versions, but as the codebase evolved, we had to add additional workarounds, like declaring imports within a class, see [channel.scala#L5-L20](https://github.com/sparklyr/sparklyr/blob/feature/sparklyr-1.0/java/spark-1.5.2/channel.scala#L5-L20).

To help others contribute with ease to `sparklyr`, this PR deprecates this embedded Livy sources connection method and enforces using the `version` parameter in Livy connections for `spark_connect()`, a feature introduced in `sparklyr 1.0`. This is also desirable since it's much faster to use Livy with a precompiled JAR than uploading sources.

New Livy connections are now expected to be initialized with the `version` parameter; otherwise, an error will be presented to the user:

```r
sc <- spark_connect(
  master = "https://domain/livy/",
  version = "2.3.1",
  method = "livy",
  config = livy_config(user = "<user>", password = "<password>"))
```

This PR does ALSO cleans up the `scala` sources, by refactoring the imports and also avoiding `invoke()` calls.


R/livy_connection.R
```diff
@@ -588,6 +593,10 @@ livy_connection <- function(master,
                             hadoop_version,
                             extensions) {
 
+  if (is.null(version)) {
+    stop("Livy connections now require a the Spark version to be specified.")
```
kevinykuo: nit: `call. = FALSE`

R/livy_connection.R
```diff
@@ -601,7 +610,7 @@ livy_connection <- function(master,
 
   livy_validate_master(master, config)
 
-  if (!is.null(version) && !spark_config_value(config, "sparklyr.livy.sources", FALSE)) {
+  if (!spark_config_value(config, "sparklyr.livy.sources", FALSE)) {
     extensions <- spark_dependencies_from_extensions(version, extensions, config)
 
     config$livy.jars <- as.character(c(livy_connection_jars(config, version), extensions$catalog_jars))
```
kevinykuo: If getting the jars from github fails for whatever reason, can we fall back to uploading from local (since I presume they're available in `inst/java`)?

javierluraschi: We are removing them completely, there is no fallback anymore with this PR! +1 -1?

R/livy_connection.R
javierluraschi: K, will update!

R/livy_connection.R
javierluraschi: Oh you mean uploading the jars... unfortunately, that's not supported by Livy :( They need to be downloaded from somewhere or manually copied into the Livy server. Would be nice!

PR #2242, opened 2020-01-16
Add sdf_drop_duplicates

zero323:
This PR adds a wrapper for `Dataset.dropDuplicates` method.

It can be used for efficient deduplication of records when particular choice of row doesn't matter, roughly equivalent to:

```r
distinct(iris, Species, .keep_all = TRUE)
```

with plain `data.frame`, which doesn't seem to play well with `sparklyr`

```r
distinct(iris_tbl, Species, .keep_all = TRUE)
## Error: Can only find distinct value of specified columns if .keep_all is FALSE
```

Additionally it has special meaning for streaming datasets.

### DIFF
--- a/R/sdf_interface.R
+++ b/R/sdf_interface.R
@@ -479,3 +479,29 @@ sdf_describe <- function(x, cols = colnames(x)) {
     invoke("describe", cols) %>%
     sdf_register()
 }
+
+#' Remove duplicates from a Spark DataFrame
+#'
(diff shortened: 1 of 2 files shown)


zero323:
Retest this please.

zero323:
Doesn't seem like Livy build failures are related to anything covered by this PR, does it?

javierluraschi:
It's not related, looks good! Thank you!

zero323:
Thanks @javierluraschi.

PR #2246, opened 2020-01-29
Basic Jenkinsfile for integration testing

Loquats:
This Jenkinsfile will be used to integration test sparklyr with Databricks Connect.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,10 @@
+pipeline {
+    agent any
+    stages {
+        stage('Stage 1') {
+            steps {
+                echo 'Hello world!'
+            }
+        }
+    }
+}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,15 +52,14 @@ pipeline {
         }
         stage("Run tests") {
             steps {
-                echo "TODO 2"
+                sh """R --vanilla --slave -e 'devtools::install(".", dependencies=TRUE, repos="https://cran.microsoft.com/snapshot/2019-04-15/")'"""
+                sh """SPARK_HOME=${sparkHome} R --vanilla --slave -e 'devtools::test()'"""
             }
         }
-        stage("Terminate Databricks cluster") {
-            steps {
-                script {
-                    sh "databricks clusters delete --cluster-id ${clusterId}"
-                }
-            }
+    }
+    post {
+        always {
+            sh "databricks clusters delete --cluster-id ${clusterId}"
         }
     }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,7 +52,7 @@ pipeline {
         }
         stage("Run tests") {
             steps {
-                sh """R --vanilla --slave -e 'devtools::install(".", dependencies=TRUE, repos="https://cran.microsoft.com/snapshot/2019-04-15/")'"""
+                sh """R --vanilla --slave -e 'devtools::install(".", dependencies=TRUE)'"""
                 sh """SPARK_HOME=${sparkHome} R --vanilla --slave -e 'devtools::test()'"""
             }
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,10 +1,46 @@
+import groovy.json.JsonOutput
+
 pipeline {
     agent any
     stages {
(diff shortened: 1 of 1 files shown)


javierluraschi:
@Loquats should this one get merged? There seem to be some integration TODOs, was that expected?

Loquats:
Hi @javierluraschi. Still working on this. I will fill out the TODO's before merging.

Jenkinsfile
```diff
@@ -0,0 +1,66 @@
...
+pipeline {
+    agent any
+    stages {
+        stage("Create Databricks cluster") {
+            steps {
+                script {
+                    def uuid = UUID.randomUUID().toString()
+                    def clusterParams = [
+                        cluster_name: "jenkins-${uuid}",
+                        spark_version: "6.2.x-scala2.11",
```
falaki: Can we parameterize this?

Jenkinsfile
```diff
@@ -0,0 +1,66 @@
...
+                script {
+                    withCredentials([
+                        string(credentialsId: 'databricks-connect-api-token', variable: 'API_TOKEN'),
+                        string(credentialsId: 'databricks-workspace-url', variable: 'WORKSPACE_URL')
+                    ]) {
+                        def dbConnectParams = [
+                            host: WORKSPACE_URL,
+                            token: API_TOKEN,
+                            cluster_id: clusterId,
+                            org_id: "0",
```
falaki: It seems we can remove this since we are using an AWS workspace.

Jenkinsfile
```diff
@@ -0,0 +1,66 @@
...
+                            port: "15001",
+                        ]
+                        def dbConnectParamsJson = JsonOutput.toJson(dbConnectParams)
+                        sh "echo '$dbConnectParamsJson' > ~/.databricks-connect"
+
+                        // Smoke test to check if databricks-connect is set up correctly
+                        def sparkHome = "/usr/lib/python3.7/site-packages/pyspark"
```
falaki: Would you make it a global variable?

falaki: Also can we set SPARK_HOME environment variable outside of this script? I am not sure which one is better just wondering.

Jenkinsfile
```diff
@@ -0,0 +1,66 @@
...
+                    }
+                }
+            }
+        }
+        stage("Run tests") {
+            steps {
+                echo "TODO 2"
+            }
+        }
+        stage("Terminate Databricks cluster") {
```
falaki: Is there a way to run this stage regardless of success or failure of previous stages? I know we are setting auto-termination timeout.

Jenkinsfile
Loquats: It's possible to set the environment variable outside of the script, for example how we define WORKSPACE_URL as a private credential string (it's probably also possible using other mechanisms but I haven't look into it). I don't know what would be the use case of defining it outside the script though.

Jenkinsfile
Loquats: Sure good idea. One minor caveat will be that it will attempt to terminate the cluster even if it wasn't successfully created in the first place, but that's fine.

Jenkinsfile
Loquats: We could, but it's probably not worth doing it at this point. The jenkins machine installs databricks-connect==6.2.0, which is only compatible with DBR 6.2, so even if we wanted to test on different runtimes, it would be a big effort, eg. to create conda environments for each DBR version.

Jenkinsfile
```diff
@@ -0,0 +1,65 @@
...
+                        // Smoke test to check if databricks-connect is set up correctly
+                        sh "SPARK_HOME=${sparkHome} databricks-connect test"
+                    }
+                }
+            }
+        }
+        stage("Run tests") {
+            steps {
+                sh """R --vanilla --slave -e 'devtools::install(".", dependencies=TRUE, repos="https://cran.microsoft.com/snapshot/2019-04-15/")'"""
+                sh """SPARK_HOME=${sparkHome} R --vanilla --slave -e 'devtools::test()'"""
```
falaki: Since travis is using CRAN, it is OK not to use MRAN snapshot here. If CRAN becomes flaky travis will fail anyway.  Plus we will see consistent behavior.

PR #2250, opened 2020-02-02
remove embedded nul bytes if present in raw data to be read

jozefhajnala:
When trying to collect data that contain embedded nul bytes into R the process fails with an error (reproducible example added as a unit test case). This proposes to work around this issue by omitting those bytes.


tests/testthat/test-read-write.R
```diff
@@ -239,5 +239,17 @@ test_that("spark_read_csv() can read verbatim column types", {
       c = list(name = "c", type = "DoubleType")
     )
   )
+})
 
+test_that("spark_read_csv() can read if embedded nuls present", {
+  fpath <- dir(getwd(), "with_embedded_nul\\.csv$", full.names = TRUE, recursive = TRUE)
```
javierluraschi:
```suggestion
  skip_on_arrow()
  
  fpath <- dir(getwd(), "with_embedded_nul\\.csv$", full.names = TRUE, recursive = TRUE)
```

javierluraschi:
Looks like `arrow` does not support this test case:

```
spark_read_csv() can read if embedded nuls present: error: embedded nul in string: 'test\0string' 
```

Skipping test for now, thank you for this fix @jozefhajnala!

jozefhajnala:
Added the skip as suggested.

falaki:
Databricks Connect tests failed. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2250-2/70f40310-c492-4ecd-a094-0731d661a2df.txt).

falaki:
Databricks Connect tests succeeded. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2250-3/719662d6-d456-42e8-b098-016b5fc45f24.txt).

falaki:
Databricks Connect tests succeeded. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2250-4/682fd9b8-a4e8-46e3-b0f3-47e4cc921b23.txt).

yl790:
Great success!

rexdouglass:
I'm encountering an error on collect
Error in RecordBatch__to_dataframe(x, use_threads = option_use_threads()) : 
  embedded nul in string: '�\0'   

I have the most updated 1.3.1 whose patch notes include "Embedded nul bytes are removed from strings when reading strings from Spark to R (#2250)" 
https://spark.rstudio.com/news/

I do not have a reproducible example (the null appears somewhere in a 13 billion row table, and creating one artificially is difficult). So i'm just asking if this hotfix is supposed to cover the use case of running a collect() on a spark table and returning it to R, as is the case where I'm getting this error.

yl790:
@rexdouglass Hey I don't think this PR covers the use case involving arrow. You can try disabling arrow as a workaround but then there will be some performance penalty as deserialization will be slower without arrow.

PR #2253, opened 2020-02-06
Support JDK11 for Spark 3

kevinykuo:
- Add support for JDK11 in Spark 3
- Add informative error message when attempting to use JDK11 with Spark 2 (manually tested only)
- Add `openjdk11` to test matrix

Closes https://github.com/sparklyr/sparklyr/issues/1922, closes https://github.com/sparklyr/sparklyr/issues/2049.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,11 @@ matrix:
       env:
         - SPARK_VERSION="master"
         - JAVA_VERSION=oraclejdk8
+    - name: "Spark master (R release, openjdk11)"
+      r: release
+      env:
+        - SPARK_VERSION="master"
+        - JAVA_VERSION=openjdk11
     - name: "Livy 0.5.0 (R release, oraclejdk8, Spark 2.3.0)"
       r: release
       env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ matrix:
       r: release
       env:
         - SPARK_VERSION="master"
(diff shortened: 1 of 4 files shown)


kevinykuo:
Looks like openjdk11 requires using `install-jdk.sh` (https://docs.travis-ci.com/user/languages/java/#using-java-10-and-later), and we also need to parse the version string for oraclejdk11.

.travis.yml
```diff
@@ -46,6 +46,11 @@ matrix:
       env:
         - SPARK_VERSION="master"
         - JAVA_VERSION=oraclejdk8
+    - name: "Spark master (R release, openjdk11)"
+      r: release
+      env:
+        - SPARK_VERSION="master"
+        - JAVA_VERSION=openjdk11
```
javierluraschi:
```suggestion
        - JAVA_URL=https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.1%2B13/OpenJDK11U-jdk_x64_linux_hotspot_11.0.1_13.tar.gz
```

We would also need to add after `before_install`:

```
- wget https://github.com/sormuras/bach/raw/master/install-jdk.sh
- if [[ ! -z "$JAVA_URL" ]]; then source install-jdk.sh --url $JAVA_URL ; fi
```

See https://github.com/danielnorberg/rut/blob/master/.travis.yml

PR #2265, opened 2020-02-27
allow doSpark to support loop referencing external functions and variables"

yl790:
Because otherwise the applicability of `doSpark` as a parallel backend for `foreach` would be severely limited. In particular, one would not be able to apply `doSpark` to the hyper parameter tuning use cases inside `tune`.

### DIFF
--- a/tests/testthat/test-do-spark.R
+++ b/tests/testthat/test-do-spark.R
@@ -11,6 +11,26 @@ register_test_spark_connection <- function() {
 
 register_test_spark_connection()
 
+u <- 1234
+v <- 5678
+x <- 123
+y <- 456
+z <- 789
+fn_1 <- function(x) {
+  list(fn_1 = list(x = x, u = u))
+ }
+fn_2 <- function(x) {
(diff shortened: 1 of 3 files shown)


tests/testthat/test-do-spark.R
```diff
@@ -11,6 +11,13 @@ register_test_spark_connection <- function() {
 
 register_test_spark_connection()
 
+y <- 3
+z <- 4
+fn <- function(x) { x * x - y }
+test_fn <- function(x) {
+  fn(fn(x * y) - z) - y * z
```
yl790: but if `test_fn` were to contain an inner function, e.g.,
```
test_fn <- function(x) {
  inner_fn <- function(x) { ... }
   ...
}
```
, then this would still break and there would be an error saying 'function "function" not found' -- I believe that may be a parsing bug from `globals`. We may need to live with this limitation for now unfortunately.

yl790:
- revised comment
- made test case cover more scenarios

PR #2266, opened 2020-02-28
wait for livy server socket to be in listening state before running test cases

yl790:
I may have figured out why the CI test script frequently fails with "Failed to launch livy session, session status is dead" even after Livy is installed successfully.

Let's see if this change helps.

### DIFF
--- a/tests/testthat/helper-initialize.R
+++ b/tests/testthat/helper-initialize.R
@@ -163,6 +163,31 @@ sdf_query_plan <- function(x) {
     unlist()
 }
 
+wait_for_svc <- function(svc_name, port, timeout_s) {
+  suppressWarnings({
+    socket <- NULL
+    on.exit(
+      if (!is.null(socket)) {
+        close(socket)
+      }
+    )
+    for (t in 1:timeout_s) {
+      try(
+        socket <- socketConnection(
+          host = "localhost", port = port, server = FALSE, open = "r+"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -96,6 +96,7 @@ matrix:
         apt:
           packages:
             - openjdk-8-jre
+            - netcat
```
yl790: actually won't need netcat anywhere... gonna wait for CI to complete, and then delete this and run CI again

javierluraschi:
@yitao-li we could also consider moving some tests to use GitHub actions... they run sooooo much faster than Travis and we've had good success in other projects like https://github.com/mlverse/lantern

javierluraschi:
BTW. Even if this doesn't fix the test, I think is a good change to merge, as in, just in case.

javierluraschi:
We might be able to get meaningful error information if we change stdout to `TRUE` here:

https://github.com/sparklyr/sparklyr/blob/master/tests/testthat/helper-initialize.R#L201-L205

We can also consider adding Livy tests to allowed failures:

https://github.com/sparklyr/sparklyr/blob/master/.travis.yml#L109-L123

javierluraschi:
My guess is that when a PR gets created, Travis uses machines with less memory, where Livy fails to start due to low memory or something around those lines.

PR #2277, opened 2020-03-03
append uuid to spark bundle filename

yl790:
Implementing the suggested change in https://github.com/sparklyr/sparklyr/issues/2244

### DIFF
--- a/R/spark_apply_bundle.R
+++ b/R/spark_apply_bundle.R
@@ -2,11 +2,11 @@ spark_apply_bundle_path <- function() {
   file.path(tempdir(), "packages")
 }
 
-spark_apply_bundle_file <- function(packages, base_path) {
+spark_apply_bundle_file <- function(packages, base_path, session_id = NULL) {
   file.path(
     base_path,
     if (isTRUE(packages))
-      "packages.tar"
+      do.call(paste, as.list(c("packages", session_id, "tar", sep = ".")))
     else
       paste(
         substr(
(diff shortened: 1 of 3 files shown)


R/spark_apply_bundle.R
```diff
@@ -6,7 +8,7 @@ spark_apply_bundle_file <- function(packages, base_path) {
   file.path(
     base_path,
     if (isTRUE(packages))
-      "packages.tar"
+      paste("packages_", uuid::UUIDgenerate(), ".tar", sep = "")
```
javierluraschi: Creating the package bundle takes a long time, so I think we should avoid creating it to only once per session.  There is a sessionId that is assigned to all connections, see https://github.com/sparklyr/sparklyr/blob/cf3f4d6e5f56856aa4f25fe887190c11b2090c8a/R/shell_connection.R#L163-L166. So I think if we reuse that we are better off to avoid slowing down subsequent calls to `spark_apply()`.

```suggestion
      paste("packages_", sc$sessionId, ".tar", sep = "")
```

R/spark_apply_bundle.R
```diff
@@ -1,3 +1,5 @@
+library(uuid)
```
javierluraschi:
```suggestion
```

DESCRIPTION
```diff
@@ -48,6 +48,7 @@ Imports:
     rstudioapi (>= 0.10),
     tibble,
     tidyr,
+    uuid,
```
javierluraschi:
```suggestion
```

PR #2284, opened 2020-03-05
Jenkins CI for Databricks Connection mode

falaki:
- Makes sure the Jenkins build fails when tests are not successful.


Loquats:
Is this all the tests that are failing? Many ML tests should be failing too, unless they've all been fixed by the latest databricks-connect pip package.

PR #2294, opened 2020-03-08
fixing the issue I found in https://github.com/sparklyr/sparklyr/issu…

yl790:
I think the bug happened because `is.character(columns)` will evaluate to true for both `columns = c(result = "double")` and `columns = c("result")` and the code path supposedly handling the latter case is also handling the former, causing a column named "double" to be created.

### DIFF
--- a/R/spark_apply.R
+++ b/R/spark_apply.R
@@ -150,6 +150,14 @@ spark_apply <- function(x,
                         ...) {
   memory <- force(memory)
   args <- list(...)
+  # If columns is of the form c("col_name1", "col_name2", ...)
+  # then leave it as-is
+  # Otherwise if it is of the form c(col_name1 = "col_type1", ...)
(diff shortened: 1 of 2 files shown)


R/spark_apply.R
```diff
@@ -150,6 +150,14 @@ spark_apply <- function(x,
                         ...) {
   memory <- force(memory)
   args <- list(...)
+  # If columns is of the form c("col_name1", "col_name2", ...)
+  # then leave it as-is
+  # Otherwise if it is of the form c(col_name1 = "col_type1", ...)
+  # or list(col_name1 = "col_type1", ...), etc, then make sure it gets coerced
+  # into a list instead of a character vector with names
```
javierluraschi: @yl790 I think is better to document these changes in the PR and briefly in the commit message. The PR can have all the details and the commit message something like `handle vector names to fix #number"`. If someone wants to understand why this got added, they can always use GitHub's blame or the associated PR. No need to update, just for future changes.

PR #2299, opened 2020-03-10
No need to set master when using `method = "databricks"`

falaki:
* User does not need to set `master = local` when using Databricks connect.
* There is a test failure which is fixed
* Adding README instructions for DB Connect.


README.md
```diff
@@ -594,3 +594,112 @@ config <- livy_config(username="<username>", password="<password>")
 sc <- spark_connect(master = "<address>", method = "livy", config = config)
 spark_disconnect(sc)
 ```
+
+
+## Connecting through Livy
```
javierluraschi: We also need this added to the `Rmd` since the `md` is generated that way.

falaki: Oh good to know. So we just update `README.Rmd` and `README.md` is auto-generated?

Issue #2292, opened 2020-03-07
AppVeyor failure in doSpark in Spark 2.4


javierluraschi:
Noticed that we are consistently getting failures in the AppVeyor test:

```
Context: do-spark
: warning: package 'foreach' was built under R version 3.6.3 
: warning: package 'iterators' was built under R version 3.6.3 
doSpark loads required packages: warning: cannot open file 'C:/Users/appveyor/AppData/Local/Temp/1\RtmpMTQVmB/working_dir\RtmpWyUucF\file6181c1b14b6_spark.log': Permission denied 
doSpark preserves exception error message: 13.2551369667053
doSpark loads required packages: error: cannot open the connection 
  callstack:
```

https://ci.appveyor.com/project/javierluraschi71148/sparklyr/build/job/hneeyxxfxlae01so/artifacts

If you ping me in Slack, Yitao, we have a shared Windows AWS machine to troubleshoot.

yl790:
@javierluraschi not entirely sure how the `(.packages())` call (to list loaded packages on workers)  generates an IO error on Windows ("cannot open the connection").

I can try the only other way I know of for determining whether a package is loaded in R: `sprintf("package:%s", pkg_name) %in% search()`. Maybe that's something that will work better on Windows somehow (?). Fingers crossed lol :crossed_fingers:

yl790:
^^ Can confirm calling `search()` instead of `(.packages())` did not help on Windows : /
So I'm looking into other possible failure reasons now.
In the worst case scenario we can always disable some test cases on Windows, but that would be non-ideal as everything in sparklyr should work on all major platforms.

yl790:
Failures seem somewhat random (as there were a small number of appveyor builds I ran that failed at other places) and seem to be related to random permission changes of temp directory on Windows.

PR #2303, opened 2020-03-10
probable workaround for windows temp dir issues

yl790:
The default Windows temp directory could be a dangerous war zone where random processes might nuke permissions or delete files (e.g., https://techcommunity.microsoft.com/t5/storage-at-microsoft/windows-10-and-storage-sense/ba-p/428270 or similar) -- So, moving the temp directory for test environment to a different user-writable directory, which fixes the spark shell log unreadable issue in #2292

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,6 +22,7 @@ environment:
     INSTALL_WINUTILS: true
     R_VERSION: release
     USE_RTOOLS: true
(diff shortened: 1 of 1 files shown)


javierluraschi:
I'm assuming the jenkins failure is unrelated since this only affects appveyor, so merging.

PR #2304, opened 2020-03-10
replace hive context with spark session for spark 3.0.0-preview

yl790:
The deprecated HiveContext class was removed from Spark 3.0.0-preview and its functionalities are available through SparkSession if enabled


yl790:
Hmm actually I think this change might no longer be needed -- The spark_submit test failure mentioned in https://github.com/sparklyr/sparklyr/issues/2193 is no longer happening with spark 3.0.0-preview

PR #2308, opened 2020-03-11
Fix unit tests that attempt to create tables that already exist

falaki:
- Randomize table names
- Clean up tables after each suite.

Fixes https://github.com/sparklyr/sparklyr/issues/2307


falaki:
Thanks @Loquats I verified that the ML suites are still failing due to other issues. We will work on those later.

PR #2309, opened 2020-03-11
Fix tests for DB connect with local data file

lu-wang-dl:
Fix sparklyr unit tests that depend on local data files.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,6 +65,7 @@ pipeline {
     }
     post {
         always {
+	    sh """dbfs rm dbfs:/tmp/data"""
             sh "databricks clusters delete --cluster-id ${clusterId}"
 	    sh """dbfs rm -r dbfs:/tmp/data"""
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -50,7 +50,7 @@ pipeline {
                 }
             }
         }
-	stage("Prepare the test data") {
+        stage("Prepare the test data") {
 	    steps {
 		sh """dbfs mkdirs dbfs:/tmp/data"""
 		sh """dbfs cp -r --overwrite tests/testthat/data dbfs:/tmp/data"""
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,7 +65,6 @@ pipeline {
     }
     post {
         always {
-	    sh """dbfs rm dbfs:/tmp/data"""
             sh "databricks clusters delete --cluster-id ${clusterId}"
 	    sh """dbfs rm -r dbfs:/tmp/data"""
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,11 +51,11 @@ pipeline {
             }
         }
         stage("Prepare the test data") {
-	    steps {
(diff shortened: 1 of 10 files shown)


Jenkinsfile
```diff
@@ -59,6 +65,7 @@ pipeline {
     }
     post {
         always {
+	    sh """dbfs rm dbfs:/tmp/data"""
```
falaki: Nit: indent.

tests/testthat/helper-initialize.R
```diff
@@ -382,3 +382,11 @@ skip_databricks_connect <- function() {
   if (Sys.getenv("TEST_DATABRICKS_CONNECT") == "true")
     skip("Test is skipped on Databricks Connect")
 }
+
+get_simple_data_path <- function(file_name) {
```
falaki: Nit: Use two spaces for indent.

falaki:
Thank you! I think you can enable the following tests too. Please give them a try.
- `test-read-write-multiple.R`
- `test-spark-utils.R`

falaki:
@Loquats are you suggesting adding new APIs for example `spark_read_dbfs()?` analogous to `spark_read_csv()` or `spark_read_json()`? Aren't these about formats rather than file locations?

tests/testthat/helper-initialize.R
```diff
@@ -382,3 +382,11 @@ skip_databricks_connect <- function() {
...
+    if (Sys.getenv("TEST_DATABRICKS_CONNECT") == "true")
+        sample_data_path <- paste("dbfs:/tmp/data/", file_name, sep="")
+    else
+        sample_data_path <- dir(dir_path, recursive = TRUE, pattern = file_name, full.names = TRUE)
```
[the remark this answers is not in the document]
javierluraschi: I think it would be useful to support `dbfs://` when using `method = "databricks"`, not sure what is missing to get this working... probably just add a few more JARs that support read/write from `dbfs://`?

I do second Hossin's comment about not adding `spark_read/write_dbfs()`, since I would expect users will want to use all the read/write functions (`spark_read_csv`, `spark_read_json`, etc) with DBFS.

lu-wang-dl:
The test `test-spark-utils.R` is passed, so I eanble it.
But the test `test-read-write-multiple.R` has the similar issue with `test-read-write.R`, we should fix these in a separate PR.

falaki:
Would you take a look at the current failure on `test-spark-utils.R`? 
```
Path must be absolute: foobar/0624d1f4-37f4-48dd-8582-95968d551b14
```

Loquats:
> @Loquats are you suggesting adding new APIs for example `spark_read_dbfs()?` analogous to `spark_read_csv()` or `spark_read_json()`? Aren't these about formats rather than file locations?

No, I agree with you and Javier. I think the best experience for the user would be being able to use existing `spark_read_*` and `spark_write_*` API's with DBFS, when sparklyr is using databricks-connect connection method. Not sure how this would be implemented. I think this is a larger task and requires a separate discussion -- just wanted to spark the conversation here.

PR #2315, opened 2020-03-12
Improve serialization and deserialization of date and timestamp types

yl790:
This PR addresses the issue in https://github.com/sparklyr/sparklyr/issues/2314

- Turns out previously 1970-01-01 00:00:00 UTC was being used as a sentinel value to represent NA in `core_deserializer.R`, which was not good because it would cause an actual timestamp / date type with value 1970-01-01 00:00:00 UTC to be non-representable after being collected from Spark to R -- A more concrete example, let's say I'm working with very long term historic data such as `ggplot2::economics`, then I could end up with employment rates from year 1968, 1969, NA, 1971, etc after collecting my date from Spark to R, which would definitely be weird. In fact, that was how I found this bug.
- It feels like the `timeNA` hack in `core_deserializer.R` was actually trying to "cancel out" a bunch of other bugs in `utils.scala`, but introduced undesired side-effects. So, this PR tries to address the acutal issue in `utils.scala` instead, and reverts the hacky change in `core_deserializer.R`.


yl790:
created another test case for "1970-01-01"

PR #2322, opened 2020-03-18
Add local property for Databricks Connect

Loquats:
This will allow DB Connect to identify whether it is being called from sparklyr (as opposed to sparkr, pyspark, etc.)

### DIFF
--- a/R/connection_spark.R
+++ b/R/connection_spark.R
@@ -274,6 +274,10 @@ spark_connect <- function(master,
     }
   }, onexit = TRUE)
 
+  if (method == "databricks-connect") {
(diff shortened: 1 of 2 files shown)


falaki:
Changes look good to me. Travis tests for Spark 1.6 are failing in master and it is due to this:
```
The command "Rscript -e 'deps <- remotes::dev_package_deps(dependencies = NA);remotes::install_deps(dependencies = TRUE);if (!all(deps$package %in% installed.packages())) { message("missing: ", paste(setdiff(deps$package, installed.packages()), collapse=", ")); q(status = 1, save = "no")}'" failed and exited with 1 during .
```

I think we are good to go on this.

PR #2324, opened 2020-03-18
fix a spark-3.0 compatibility issue when connecting to an existing spark session

yl790:
This PR should fix the issue described in https://github.com/sparklyr/sparklyr/issues/2321

### DIFF
--- a/R/shell_connection.R
+++ b/R/shell_connection.R
@@ -566,50 +566,57 @@ shell_connection_config_defaults <- function() {
 
 #' @export
 initialize_connection.spark_shell_connection <- function(sc) {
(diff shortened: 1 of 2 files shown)


javierluraschi:
Nice fix! There is a simpler one but hard to known unless someone explains that we have a `devtools` package to install for GitHub, instead of patching the RCpp headers we can also do something like this:

https://github.com/rstudio/pins/pull/192/commits/10feae4c862157f110f8d4740ea75fd1402082a5

My guess is that this will be patch in CRAN soon, so we can merge as it is and revert later on.

PR #2325, opened 2020-03-18
[experimental][do-not-merge] try to run some tests in parallel

yl790:
Looks like some tests might fail for various random reasons, even though each one is spawning its own separate backend channel. Probably worth investigating further. Maybe it's just a matter of closing backend channels after the process is done (? but I thought they were auto-closed)

It would be nice if this just works for all tests and helps us speed up CI build


javierluraschi:
Spark is unfortunately not multi-threaded, since it's already parallelizing everything anyways. We could parallelize the R code, but I don't think that's too heavy. Probably the best way to improve the tests here is to actually make sparklyr faster. You could try using `profvis` when running the `testthat` tests, see (https://support.rstudio.com/hc/en-us/articles/218221837-Profiling-with-RStudio), and see if you can improve execution time across sparklyr. My guess is that `invoke()` calls might be our best case to speed up this. In any case, glad you are taking a look at this!

PR #2329, opened 2020-03-19
[do-not-merge-yet] try running tests without stringi package

yl790:
The rcpp dependency from stringi stopped compiling with R 3.2 headers and I can't find a reasonable workaround for that.

We can see which tests would break without stringi (hopefully few to none) and then just skip those tests if R version is 3.2 or below.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ cache:
 
 warnings_are_errors: true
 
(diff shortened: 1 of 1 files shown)


yl790:
Turns out the rcpp compilation errors were unrelated to stringi

PR #2332, opened 2020-03-19
workaround for Rcpp 1.0.4 compilation failure with R 3.2

yl790:
This compilation error has been blocking one of the CI jobs recently. The only solution I can think of is creating a bash script that conditionally patches the Rinternals.h, which is wonderfully insane, but lo and behold it worked!

(note: this is only OK because only some R package travis uses has dependency on Rcpp 1.0.4. Sparklyr currently does not depend on Rcpp 1.0.4 in any way)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,6 @@ branches:
 cache:
   packages: true
 
-warnings_are_errors: true
-
 r_github_packages:
   - gagolews/stringi@v1.4.2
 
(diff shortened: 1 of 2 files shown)


javierluraschi:
Nice fix! There is a simpler one but hard to known unless someone explains that we have a `devtools` package to install for GitHub, instead of patching the RCpp headers we can also do something like this:

https://github.com/rstudio/pins/pull/192/commits/10feae4c862157f110f8d4740ea75fd1402082a5

My guess is that this will be patch in CRAN soon, so we can merge as it is and revert later on.

PR #2341, opened 2020-03-23
optimize jvm object deletions

yl790:
- [TODO] avoid waiting for objects to be removed from jvm object tracker (as those are triggered by gc anyways and no other code path should depend on the completion of those operations)
- remove a list of objects using 1 invoke call


yl790:
This change only shortened the CI build times by 1-2 mins though. There are probably some other bigger wins to be had from elsewhere.

PR #2346, opened 2020-03-25
Allow sdf_query_plan to also get analyzed plan

Loquats:
Allow sdf_query_plan to also get analyzed plan, which is more useful for DB Connect. The PR adds an optional argument to `sdf_query_plan`, to get either the optimized or analyzed plan. When called with DB Connect, the optimized plan is just a LogicalRemoteExec "black box", and doesn't display much useful query information. Instead, the analyzed plan can be used to examine the query. 

For backwards compatibility, the `plan_type` argument of `sdf_query_plan` defaults to `optimizedPlan`, so that existing usage will have the same behavior.

This also fixes and enables two tests for DB Connect that check query plans.

### DIFF
--- a/tests/testthat/helper-initialize.R
+++ b/tests/testthat/helper-initialize.R
@@ -158,11 +158,13 @@ check_tidy <- function(o, exp.row = NULL, exp.col = NULL, exp.names = NULL) {
   }
 }
 
-sdf_query_plan <- function(x) {
+sdf_query_plan <- function(x, plan_type=c("optimizedPlan", "analyzed")) {
+  plan_type <- match.arg(plan_type)
+
   x %>%
     spark_dataframe() %>%
     invoke("queryExecution") %>%
-    invoke("optimizedPlan") %>%
+    invoke(plan_type) %>%
     invoke("toString") %>%
     strsplit("\n") %>%
     unlist()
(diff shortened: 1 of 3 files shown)


tests/testthat/helper-initialize.R
```diff
@@ -158,11 +158,13 @@ check_tidy <- function(o, exp.row = NULL, exp.col = NULL, exp.names = NULL) {
   }
 }
 
-sdf_query_plan <- function(x) {
+sdf_query_plan <- function(x, plan_type=c("optimizedPlan", "analyzed")) {
+  plan_type <- match.arg(plan_type)
+
```
Loquats: Optional argument defaults to `optimizedPlan` for backwards compatibility.

tests/testthat/test-partitioning.R
```diff
@@ -10,24 +10,34 @@ test_that("sdf_repartition works", {
...
-                 gsub("#[0-9]+", "", ., perl = TRUE),
-               "RepartitionByExpression [Species, Petal_Width], 5")
+  expect_true(
+    any(grepl("RepartitionByExpression \\[Species, Petal_Width\\], 5",
+              iris_tbl %>%
+                sdf_repartition(5L, partition_by = c("Species", "Petal_Width")) %>%
+                sdf_query_plan(plan_type) %>%
+                gsub("#[0-9]+", "", ., perl = TRUE))
+    )
+  )
```
Loquats: The new assertions are more general, and a bit looser, but I don't expect any issues to show up, since the string that we are grepping for is very specific.

tests/testthat/test-dplyr.R
```diff
@@ -122,13 +122,20 @@ test_that("'sample_n' and 'sample_frac' work in nontrivial queries (#1299)", {
 })
 
 test_that("'sdf_broadcast' forces broadcast hash join", {
-  skip_databricks_connect()
+  if (sc$method == "databricks-connect") {
```
falaki: For consistency isn't it better to use `is_testing_databricks_connect()` instead of `sc$method == "databricks-connect"`?

falaki:
@yl790 it seems workflows are broken by this commit https://github.com/sparklyr/sparklyr/commit/704963b621abf6109a2f2c27de80ce89dd68365c

falaki:
@javierluraschi and @yl790 what are the plans for CI? Are we moving off travis?

yl790:
Hi Hossein:
    Yes I'm exploring the possibility of migrating from travis to github action. I'll let you know how it goes. Meanwhile I should probably revert the dummy github action ci.yaml commit because it's not useful yet.

Yitao

falaki:
@yl790 Thank you!

PR #2348, opened 2020-03-25
merge a chain of invoke(...) calls into a single invoke(..., "%>%", ...) call

yl790:
This change should help us avoid shuffling between R runtime and JVM while invoking a chain of JVM methods, which should reduce number of context switches, page faults, etc


javierluraschi:
This change is quite great, @yl790! I wonder if there are more places where we can take advantage of this!

We should also add entries to the [NEWS](https://github.com/sparklyr/sparklyr/blob/master/NEWS.md) file with each improvement, we will probably have to go back to all your PR's and update the NEWS file to communicate all the great improvements you've added to this release.

yl790:
Travis CI build doesn't show up here any more for some reason... I had to trigger a manual build for this commit in https://travis-ci.org/github/yl790/sparklyr/builds/666921247

PR #2354, opened 2020-03-27
setup github CI workflow

yl790:
- Finally got the github CI config mostly working
- But there are some improvements to be made such as querying R-release and oldrel versions rather than hard-coding them, caching Arrow build artifacts to avoid re-building Arrow, etc
- Still need to look into a compilation error with Arrow 0.11.0 (ref = dc5df8f)


yl790:
need to do some git house keeping and re-submit this PR

PR #2356, opened 2020-03-27
github CI workflow

yl790:
- Finally got the github CI config mostly working. Had to make some minor adjustments to test-print.R
- There are still some improvements to be made such as querying R-release and oldrel versions rather than hard-coding them, caching Arrow build artifacts to avoid re-building Arrow, etc
- Still need to look into a compilation error with Arrow 0.11.0 (ref = dc5df8f)

### DIFF
--- a/ci/arrow-build.sh
+++ b/ci/arrow-build.sh
@@ -2,21 +2,26 @@ ARROW_VERSION=$1
 
 pushd ..
 
+TMP_DIR=/tmp
+BUILD_DIR="$TMP_DIR/apache-arrow-$ARROW_VERSION-build"
+
+cd "$TMP_DIR"
+
 sudo apt install -y -V cmake
 sudo apt install -y -V libboost-all-dev
 sudo apt install -y -V autoconf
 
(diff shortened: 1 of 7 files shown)


ci/arrow-build.sh
```diff
@@ -32,9 +37,9 @@ cmake -DARROW_BUILD_TESTS=FALSE \
       -DARROW_FILESYSTEM=ON \
       -DARROW_JSON=ON \
       -DARROW_PARQUET=ON \
-      ..
+      "$SRC_DIR"
 
-make arrow
+make -j 4 arrow
```
javierluraschi: `-j 4` ^_^

javierluraschi: Oh I guess, should this be an environment variable provided by GitHub actions or do we know it's usually 4?

yl790: Yeah on a second thought probably makes sense to replace this with `$(nproc)` instead, which would be slightly more future-proof :)

PR #2361, opened 2020-03-30
comment out CI builds that are migrated to github action workflow

yl790:
- Unfortunately there are 2 CI builds (Arrow 0.11.0 & Arrow 0.13.0) that cannot be migrated to the new github action workflow yet
- Both ran into weird compilation errors either with thrift dependency or with Rcpp if building them with existing build scripts in the ci/ directory at dependency installation stage and all the workarounds I could think of did not fix those errors
- Meanwhile Travis is just downloading already compiled artifacts from its internal cache and running everything without compiling any source code at all, which is fine... but that means it won't offer us any insight whether something could be done differently to fix the compilation errors with github action workflow.
 - Also now that the Livy test is re-enabled, there is apparently some test failure that needs to be addressed there (this should be simple, but will be in another PR)

So, leaving those 2 in .travis.yml for now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,51 +12,51 @@ r_github_packages:
 
 matrix:
   include:
-    - name: "Spark 1.6.3 (R 3.2, openjdk7)"
-      r: 3.2
-      env:
-        - SPARK_VERSION="1.6.3"
-        - JAVA_VERSION=openjdk7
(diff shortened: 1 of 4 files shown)


yl790:
Hmm need to look into random test failures that are definitely unrelated to this change -- the Livy test failure is a known issue (minor) but I have no idea about the others... Everything was working fine last Friday lol

PR #2366, opened 2020-03-30
fix serialization test failure

yl790:
For some unknown reason, this test started failing recently with a type mismatch error:

 error: No common type for `svalue` <character> and `svalue` <double> and is failing on both Travis and Github CI machines.

I have no idea why it only started failing recently. But the best guess off the top of my head is this change should fix it.


yl790:
This change appears to have fixed vast majority of test failures, except for the 4 other unknown failures with Spark 1.6.3 and the known Livy issues being addressed in another PR.

PR #2369, opened 2020-03-31
Add the dockerfile for jenkins server

lu-wang-dl:
Add a Dockerfile for installing Jenkins. You should build with
```
sudo docker build .
```
Then run with
```
sudo docker run --rm --detach --network jenkins --env DOCKER_HOST=tcp://docker:2376 --env DOCKER_CERT_PATH=/certs/client --env DOCKER_TLS_VERIFY=1 --publish 8080:8080 --publish 50000:50000 --volume jenkins-data:/var/jenkins_home  --volume jenkins-docker-certs:/certs/client:ro --name jenkins-blueocean
```


ci/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM jenkinsci/blueocean
```
falaki, line 1: Would you add some inline comments to explain what this docker container is used for for the community?

PR #2374, opened 2020-04-02
support adls credential passthrough for databricks connections

yl790:
I think in theory this should fix the issue in https://github.com/sparklyr/sparklyr/issues/2342
I'm just not sure what's the easiest way to test such change E2E.

### DIFF
--- a/R/databricks_connection.R
+++ b/R/databricks_connection.R
@@ -68,6 +68,11 @@ new_databricks_connection <- function(scon, guid) {
   session <- invoke_static(sc, r_driver_local, "getDriver", guid) %>%
     invoke("sqlContext") %>%
     invoke("sparkSession")
+  # Apply user-specified fs.azure.* config values
+  fs.azure.prefix <- "fs.azure."
+  fs.azure.configs <- connection_config(sc, fs.azure.prefix)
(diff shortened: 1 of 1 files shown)


stephlocke:
Unfortunately, this didn't fix the issue

yl790:
@stephlocke  OK good to know... So the fs.azure.* config values are probably getting dropped yet again elsewhere.

I'm a bit busy with releasing sparklyr-1.2.0 at the moment. I'll try look more into this when I have time.

stephlocke:
Much appreciated @yl790 !

falaki:
Thanks for doing this. 

@stephlocke did you test this inside a Notebook or using Databricks Connect?

falaki:
> Change looks reasonable, +1 for me since user confirmed the fix works properly. Would be nice to have a thumbs up from Databricks before merging (@falaki, @lu-wang-dl) but otherwise, I think we want this fix in the CRAN release.

Sorry I think I missed this and the release is over. I would follow up with this user because there are several discussions about ADLS passthrough especially with Databricks Connect.

yl790:
Alright gonna merge this change then, because I think it's definitely a *prerequisite* for getting ADLS passthrough working.

PR #2392, opened 2020-04-08
make sdf_copy_to map list columns to Spark SQL structs

yl790:
This PR implements the feature proposed in https://github.com/sparklyr/sparklyr/issues/1615


yl790:
Note to myself: ideally StructTypes should also be deserialized into lists / named lists in sdf_collect (which I imagine should not be hard to implement after this change is finalized)

PR #2399, opened 2020-04-09
support printing spark context

yl790:
As described in https://github.com/sparklyr/sparklyr/issues/1944 -- seems to be an easy debuggability win


DESCRIPTION
```diff
@@ -70,3 +70,214 @@ Suggests:
     reshape2,
     shiny (>= 1.0.1),
     testthat
+Collate:
```
javierluraschi, line 73: Why do we need to add the explicit files? We try to avoid doing this in R packages since it makes it annoying to add new files.

yl790: Oh I think this was because I wanted to have

`print_jobj.livy_connection <- print_jobj.spark_shell_connection`

in `livy_connection.R` similar to what `spark_gateway.R` had, but that wouldn't work if the .R files are loaded in alphabetical order (as they currently do), and then fixing the file loading order with `@include` directive triggered this.

yl790: but anyhow, looks like when a new file is added, roxygen will automatically figure out where to put that new file in this list, which is awesome

javierluraschi: Ahhh, gotcha, thanks for the explanation!

PR #2400, opened 2020-04-09
call enableHiveSupport in SparkSession builder

yl790:
https://github.com/sparklyr/sparklyr/issues/1917 reminds me `enableHiveSupport` should have been called on `SparkSession` builder (even though it's probably enabled by default anyways)


yl790:
There is some weird spark-submit failure with spark-3.0.0-preview that is fixed with spark-3.0.0-preview2. So I'll have another small change ensuring all other test cases are passing after switching spark-master to spark-3.0.0-preview2, and then rebase and merge this change afterwards.

yl790:
Tests finally passing!

PR #2404, opened 2020-04-10
deserialize Spark SQL StructTypes back to sensible R data structures

yl790:
I think this would be part of what is required from https://github.com/sparklyr/sparklyr/issues/1628


yl790:
Update: there was an interesting backward compatibility issue with Spark 2.2.1 from the previous rev, but that's definitely fixed now.

PR #2407, opened 2020-04-10
Fixing Databricks CI

falaki:
Databricks tests are failing `test-spark-apply-bundle.R`.


falaki:
@yl790 would you please take a look? This commit broke Databricks CI tests: https://github.com/sparklyr/sparklyr/commit/7c8e80116dbabc30ae205f4db543581650da8036

Loquats:
Thanks for taking care of this!

falaki:
@javierluraschi and @yl790 are you OK with this change?

yl790:
@falaki Looks good to me!
I still don't think (at least in principle) this test should be tar-ing everything that is within `.libPaths()`, but after reducing the number of dependencies needed for the CI environment, this could still be OK (while being the simplest way of making sure Databricks CI and other CI workflows are all happy).

PR #2412, opened 2020-04-14
update feature/sparklyr 1.2.0 to pick up some recent minor changes

yl790:
mainly picking up the bug fix in https://github.com/sparklyr/sparklyr/issues/2240 so that `packages = "delta"` will work with spark 3.0.0-preview


yl790:
will create a new branch instead...

PR #2413, opened 2020-04-14
creating release branch for sparklyr 1.2.0

yl790:


falaki:
@yl790 thanks for starting the release branch. I see the results of code snippets under `Databricks Connect` are missing. Do you need help with running the Rmd file for that section?

yl790:
Hi, Hossein:
    Good catch! I think it means I need to re-run that step on a databricks
cluster. I'll try that ASAP and see how it goes.
    Thanks!
Yitao

falaki:
Actually you just need to run it with Databricks Connect configured on your machine. Here are instructions: https://docs.databricks.com/dev-tools/databricks-connect.html

You already have access to a Databricks Workspace that you can use. @javierluraschi can probably add you to the workspace if you do not have it.

PR #2479, opened 2020-05-11
include embedded R source as resource within sparklyr jar file

yl790:
Eventual goals are the following:

- We should be able to update the embedded R source without having to re-compile any of the scala files
- Also, be able to easily check whether the embedded source matches `R/*worker*.R` and `R/*core*.R` as part of the CI workflow


falaki:
Databricks Connect tests failed. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2479-1/e3191943-2982-4ed3-9a26-376a991d7f1d.txt).

falaki:
Databricks Connect tests failed. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2479-2/85ed6d7d-8699-4f4f-8b56-43b13bee5f1f.txt).

yl790:
Hmm I think this change even sped up the scala soures compilation time quite a bit because it got rid of the long string literal in the `sources.scala` file. This was an unexpected additional bonus lol : D Dunno how exactly scala compiler works and why the long string literal that previously existed in `sources.scala` slowed it down so much (is this all because parsing took that much longer?? which I find hard to believe), but anyways happy to report compilation is noticeably faster now.

PR #2482, opened 2020-05-12
create scripts to efficiently update and verify embedded sources

yl790:
- This means modifying the embedded R source file is now completely decoupled from compiling the scala source files, as it should be.
- Also, there is now a script to quickly verify the embedded R source file within each sparklyr jar file matches the content of `R/*{core,worker}*.R` files within the repository. This is an important sanity-check given how the embedded R source file could otherwise be out-of-sync with the rest of the repo without the discrepancy being noticeable in any way.


falaki:
Databricks Connect tests succeeded. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2482-1/c66828ea-8ec4-419d-be05-cea253b0a245.txt).

PR #2491, opened 2020-05-13
wait for backend to fully shutdown in spark_disconnect

yl790:
should wait for backend to stop accept new connections before returning


yl790:
this is not the right fix

PR #2568, opened 2020-06-22
avoid distributing jar files that are available in maven central

yl790:
Because those jar files typically should be downloaded from maven central once on client side and then cached forever in ~/.ivy2 or similar


falaki:
Databricks Connect tests succeeded. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2568-1/1bbfef7d-4e5b-4a76-b8a6-6e3696e870ef.txt).

PR #2659, opened 2020-08-13
avoid spark_read_compat_param name collision

yitao-li:
Append uuid to name <- basename(path) in `spark_read_compat_param()` so that reading from 2 different paths with the same basename does not cause one resulting data frame to overwrite the other

### DIFF
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -1,29 +1,28 @@
 #' @include avro_utils.R
 #' @include spark_apply.R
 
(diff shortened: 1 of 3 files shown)


falaki:
Databricks Connect tests failed. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2659-1/dddad01f-4c76-45c5-8585-8deacfd99d24.txt).

PR #2684, opened 2020-09-03
replace `TABLESAMPLE` with SamplingUtils.sampleWith{,out}Replacement

yitao-li:
There were a number of issues with `TABLESAMPLE`:

- The number of rows returned from it appears to be non-deterministic without a `REPEATABLE` clause, and even with a `REPEATABLE` clause, the actual fraction of rows being returned is still usually somewhat off from the expected fraction for some reason
- I don't think `TABLESAMPLE` performs sampling-with-replacement, so, the previous implementation which ignored the `replace` parameter seems to be technically incorrect if called with `replace = TRUE`


falaki:
Databricks Connect tests succeeded. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2684-1/326e6438-edda-44bf-8315-52ddacdcd132.txt).

falaki:
Databricks Connect tests succeeded. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2684-3/5a5ce5ac-5548-4236-9948-6fbeaaccd14a.txt).

PR #2767, opened 2020-10-29
fix sdf_seq() repartition default value

### LINKED ISSUE #2765: sdf_seq() error: `repartition` cannot be casted to an integer vector
`sparklyr::sdf_seq()` will not work with the default `repartition` parameter. This is because it gets passed `type` which by default is one of `"integer"` or `"integer64"`, a character string.

```r
sparklyr::sdf_seq(sc)                                                                                                                                                       
# Error: `repartition` cannot be casted to an integer vector.
```

Also, "casted" is not a word; it should say "cast" - though this is a problem for the forge package.

Looking into the code, `repartition` gets passed to `cast_scalar_integer()` which in turn passes it to `case_integer()`.

```
r$> devtools::session_info()                                                                                                                                                  
Session info ------------------------------------------------------------------------------------------------------------------------------------
 setting  value                       
 version  R version 3.4.4 (2018-03-15)
 system   x86_64, linux-gnu           
 ui       vscode                      
 language (EN)                        
 collate  en_US.UTF-8                 
 tz       Etc/UTC                     
 date     2020-10-28

Packages ------------------------------------------------------------------------------------------------------------------------------------
 package       * version    date       source   
 forge           0.2.0      2019-02-26 CRAN (R 3.4.4)
 sparklyr        1.3.0      2020-06-27 url
```

Spark 2.4.4

nathaneastwood:
Fixes #2765

FWIW, when I ran `devtools::document()`, it also updated a few of the docs but I left those out of this PR since they were out of scope to the problem.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: sparklyr
 Type: Package
 Title: R Interface to Apache Spark
-Version: 1.4.0
+Version: 1.4.0.9000
 Authors@R: c(
     person("Javier", "Luraschi", email = "javier@rstudio.com", role = "aut"),
     person("Kevin", "Kuo", role = "aut", email = "kevin.kuo@rstudio.com",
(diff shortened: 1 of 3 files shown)


yitao-li:
@nathaneastwood That's perfect. Thanks for contributing to `sparklyr`!! 👍

PR #2794, opened 2020-11-10
ensure `copy_to()` and `collect()` preserve `NA_real_` correctly

yitao-li:
After the re-write of `sparklyr` serialization routines, `copy_to()` and `collect()` can now preserve `NA_real_` correctly: `NA_real_` in R becomes `null` in Scala, which gets collected back to R as `NA_real_` (instead being reported incorrectly as `NaN` like it previously was)

Implementation-wise, there is now a custom `Numeric` case class in Scala that maps 1-1 to the `numeric` type in R. It's implemented as a `Option[Double]` with `None` mapping to `NA_real_`, `Some(NaN)` mapping to `NaN`, `Some(<number>)` mapping to `<number>`, etc.


falaki:
Databricks Connect tests succeeded. View logs [here](https://sparklyr-jenkins.s3.amazonaws.com/logs/jenkins-sparklyr-PR-2794-1/62216e1d-f07c-4b86-aeb2-0830015d7154.txt).

PR #2955, opened 2021-03-01
correct arrow_enabled to match usage

andrew-christianson:
Callers of arrow_enabled pass a spark connection, but the implementaion
of that function tries to read a config value from its first argument.
To match with its usage, this commit changes arrow_enabled to read a
config value off of sc$config.


yitao-li (approved):
Good catch! 👍

PR #2958, opened 2021-03-04
Fix transmute() such that calculations aren't performed twice

nathaneastwood:
This fixes tidyverse/dbplyr#605

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: sparklyr
 Title: R Interface to Apache Spark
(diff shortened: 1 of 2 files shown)


yitao-li:
@nathaneastwood Hey thanks for investigating this sparklyr issue.
I'm still recovering from vacation (sort of), will look into this asap.
I think this PR may be very close to fixing the problem, but it breaks some tests at the moment, so, I will need to modify it in some way before it can be merged.

nathaneastwood:
Sure, no problem. Let me know if you need anything from me.

PR #2968, opened 2021-03-09
Add sdf_distinct()

jozefhajnala:
Following the discussion in https://github.com/tidyverse/dbplyr/issues/611, I thought it would be useful to allow a somewhat more direct access to the [`distinct` Dataframe method](https://spark.apache.org/docs/1.5.1/api/java/org/apache/spark/sql/DataFrame.html#distinct()).

I still tried to keep the UX as close to the "dplyr experience" as possible, see unit tests for examples.


yitao-li (approved):
Looks great :slightly_smiling_face: 
Thanks for contributing to sparklyr!! :+1:

jozefhajnala:
For your consideration: Kind of silly of me, but now I noticed that `sdf_drop_duplicates()` is already implemented (on the Scala side I believe `distinct` is exactly the same as `dropDuplicates`) - albeit with a slightly different UX approach (argument structure on the R side)

PR #2974, opened 2021-03-15
make dbExistsTable perform case-insensitive comparisons on table names

yitao-li:


nathaneastwood:
I think this fix might cause issues in instances where Spark is case-sensitive?

yitao-li:
@nathaneastwood Are you referring to instances where `spark.sql.caseSensitive` is set to "true"? I thought that configuration only applies to column name resolutions.

nathaneastwood:
Ah, it's possible. Sorry, I'm not a Spark expert :D

nathaneastwood:
Awesome 👍🏼

PR #2983, opened 2021-03-22
Update tidyr_fill.R

mzorko:
The function is applying fill to all columns (cols) and not only on selected variables (vars).


yitao-li:
@mzorko Great catch. Thanks for contributing to sparklyr!

yitao-li (approved):

PR #3027, opened 2021-04-09
increase wait duration for livy server shutdown

yitao-li:


yitao-li:
I don't think this fixes the problem.

PR #3098, opened 2021-06-02
warn user if invalid or non-existent config file is provided to spark_config()

jozefhajnala:
Currently if the user provides an invalid or non-existent config `file` argument to `spark_config()`, that file is silently ignored. This PR proposes to warn the user in case `file` is explicitly provided but will be ignored. The added unit test cases show 2 common scenarios in which the warning can be useful.


yitao-li (approved):
Good catch! :+1: Thanks for contributing to `sparklyr`.

PR #3106, opened 2021-06-04
preserve NaN correctly when serializing Spark invoke args

yitao-li:


R/core_serialize.R
```diff
@@ -47,7 +47,7 @@ writeObject <- function(con, object, writeType = TRUE) {
   type <- class(object)[[1]]
 
   if (type %in% c("integer", "character", "logical", "double", "numeric", "factor", "Date", "POSIXct")) {
-    if (is.na(object)) {
+    if (is.na(object) && !is.nan(object)) {
```
yitao-li, line 50: TODO: create at least 1 test case for this. For example:

```
library(sparklyr)

print(invoke_static(sc, "sparklyr.Test", "infer", NaN))
```

Previous output: `"String"` (incorrect)
Expected output (which was indeed the case after this change): `"Double"`

PR #3115, opened 2021-06-08
adding missing null checks + other bug fixes and simplifications for array collectors

yitao-li:


java/spark-1.5.2/collectors.scala
```diff
@@ -30,7 +30,11 @@ object Collectors {
   def collectBooleanArr(row: Row, idx: Int): Array[Int] = {
     val arr = row(idx).asInstanceOf[scala.collection.mutable.WrappedArray[_]]
 
-    arr.map(extractBoolean).toArray
+    if (arr == null) {
```
yitao-li, line 33: TODO: should add unit tests covering each of those cases as existing unit tests don't seem to cover any of those : /

yitao-li: Done.

Issue #3140, opened 2021-06-30
Can not set "livy.rsc.server.connect.timeout"


xuzikun2003:
# Reporting an Issue with sparklyr

Here is the connection code:
s_config <- spark_config()
s_config["livy.rsc.server.connect.timeout"] <- 600
config <- livy_config(s_config, "***username***", "***pwd***")
sc <- spark_connect(master = "***url***", method="livy", config=config)

Here is the error:
Error in livy_validate_http_response("Failed to create livy session",  :
  Failed to create livy session (Client error: (400) Bad Request): "Unrecognized field \"rsc.server.connect.timeout\" (class org.apache.livy.server.interactive.CreateInteractiveRequest), not marked as ignorable (15 known properties: \"executorCores\", \"conf\", \"driverMemory\", \"name\", \"driverCores\", \"pyFiles\", \"archives\", \"kind\", \"files\", \"jars\", \"proxyUser\", \"numExecutors\", \"heartbeatTimeoutInSecond\", \"queue\", \"executorMemory\"])\n at [Source: (org.eclipse.jetty.server.HttpInputOverHTTP); line: 1, column: 57] (through reference chain: org.apache.livy.server.interactive.CreateInteractiveRequest[\"rsc.server.connect.timeout\"])"
Calls: spark_connect ... tryCatch -> tryCatchList -> tryCatchOne -> <Anonymous>
In addition: Warning messages:
1: In value[[3L]](cond) :
  Failed to create Livy session. Retrying in TRUE second(s)
2: In value[[3L]](cond) :
  Failed to create Livy session. Retrying in TRUE second(s)
3: In value[[3L]](cond) :
  Failed to create Livy session. Retrying in TRUE second(s)
Execution halted

yitao-li:
OK. I guess this is resolved now. Thanks for your PR 👍

PR #3141, opened 2021-06-30
Unblock the use of Livy client config starting with livy.rsc

xuzikun2003:
Unblock the use of the Livy client config starting with "livy.rsc". See issue #3140


yitao-li (approved):
Looks good. Thanks for contributing to sparklyr!

PR #3147, opened 2021-07-09
Implement sparklyr changes for Databricks notebook-scoped libraries

jlee2000:
I am creating this PR to give sparklyr users a better experience on Databricks with respect to notebook-scoped libraries.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -76,6 +76,12 @@
 - Fixed a bug that previously caused invoke params containing `NaN`s to be
   serialized incorrectly.
 
+### Connections
+
+- Added support for notebook-scoped libraries on Databricks connections.
+  Previously, libPaths were not shared between driver and worker in sparklyr.
+  Now, they are.
+
 ### Spark ML
 
 - `ml_compute_silhouette_measure()` was implemented to evaluate the
--- a/R/databricks_connection.R
+++ b/R/databricks_connection.R
@@ -40,6 +40,11 @@ databricks_connection <- function(config, extensions) {
     }
   )
 
+  # Hand in driver's libPaths to worker if user hasn't already set libpaths for notebook-scoped libraries
+  if (is.null(config$spark.r.libpaths)) {
+    config$spark.r.libpaths <- toString(.libPaths())
+  }
+
   new_databricks_connection(
     gateway_connection(
       paste("sparklyr://localhost:", gatewayPort, "/", gatewayPort, sep = ""),
--- a/R/databricks_connection.R
+++ b/R/databricks_connection.R
@@ -41,9 +41,7 @@ databricks_connection <- function(config, extensions) {
   )
 
   # Hand in driver's libPaths to worker if user hasn't already set libpaths for notebook-scoped libraries
-  if (is.null(config$spark.r.libpaths)) {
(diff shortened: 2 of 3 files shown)


R/databricks_connection.R
```diff
@@ -40,6 +40,11 @@ databricks_connection <- function(config, extensions) {
     }
   )
 
+  # Hand in driver's libPaths to worker if user hasn't already set libpaths for notebook-scoped libraries
```
falaki, line 43: Can we add a unit test for this in https://github.com/sparklyr/sparklyr/blob/main/tests/testthat/test-databricks-connect.R?

R/databricks_connection.R
```diff
@@ -40,6 +40,11 @@ databricks_connection <- function(config, extensions) {
...
+  if (is.null(config$spark.r.libpaths)) {
+    config$spark.r.libpaths <- toString(.libPaths())
```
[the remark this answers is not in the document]
yitao-li: We also typically shorten the `if (is.null(x)) x <- expr` type of assignment into `x <- x %||% expr` in `sparklyr` code base using the `%||%` operator from `rlang`.

yitao-li (changes requested):
Looks good. :+1: Thanks for contributing to `sparklyr`!

- I hope this will address some pain points with Databricks use cases. Do you want to also update the `NEWS.md` file outlining the effect of this change, and also (maybe) briefly the rationale behind it?

- Also a question I have: does this make sense for all Databricks use cases? Are directory layouts on Spark driver and Spark worker noes guaranteed to be identical for notebook-scoped libs?

- Please see inline comment for suggestions on `toString()` and coding style in R.

R/databricks_connection.R
```diff
@@ -40,6 +40,11 @@ databricks_connection <- function(config, extensions) {
...
+  }
```
yitao-li, line 46:
```suggestion
    config$spark.r.libpaths <- config$spark.r.libpaths %||% paste(.libPaths(), collapse = ",")
```

yitao-li: We also typically shorten the

```
if (is.null(x))
  x <- expr
```
type of assignment into `x <- x %||% expr` in sparklyr code base using the `%||%` operator from rlang.

yitao-li (changes requested):

R/databricks_connection.R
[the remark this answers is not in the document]
jlee2000: You're right, that works much better.

jlee2000:
@yitao-li Thank you for the comments! They were very helpful. To answer your question: yes, this makes sense for Databricks use cases. The directory layout for workers and drivers is guaranteed to be the same.

NEWS.md
```diff
@@ -76,6 +76,12 @@
 - Fixed a bug that previously caused invoke params containing `NaN`s to be
   serialized incorrectly.
 
+### Connections
+
+- Added support for notebook-scoped libraries on Databricks connections.
```
falaki, line 81:
```suggestion
- Added support for notebook-scoped libraries on Databricks.
```

falaki (approved):

yitao-li (approved):
LGTM :slightly_smiling_face:

Issue #1383, opened 2018-03-31
Java version detected but couldn't parse version from: java version "10" 2018-03-20


grantog:
# Reporting an Issue with sparklyr

I continue to get this error despite trying different versions of Spark and Java. Any help would be appreciated.

```r
library(sparklyr)
library(dplyr)


options("java.home"="/Library/Java/JavaVirtualMachines/jdk-10.jdk/Contents/Home/lib")
Sys.setenv(LD_LIBRARY_PATH='$JAVA_HOME/server')
dyn.load('/Library/Java/JavaVirtualMachines/jdk-10.jdk/Contents/Home/lib/server/libjvm.dylib')
library(rJava)
sc <- spark_connect(master = "local")
```
Returns:
```
* Using Spark: 2.2.0
Error in validate_java_version_line(master, version) : 
  Java version detected but couldn't parse version from: java version "10" 2018-03-20
```

Output of `utils::sessionInfo()` results in:
```
> utils::sessionInfo()
R version 3.4.2 (2017-09-28)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS High Sierra 10.13.4

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.4/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] rJava_0.9-9         sparklyr_0.7.0-9014

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.15     dbplyr_1.2.1     compiler_3.4.2   pillar_1.1.0     plyr_1.8.4       bindr_0.1        base64enc_0.1-3  tools_3.4.2      digest_0.6.15   
[10] memoise_1.1.0    jsonlite_1.5     tibble_1.4.2     nlme_3.1-131     lattice_0.20-35  pkgconfig_2.0.1  rlang_0.2.0      psych_1.8.3.3    shiny_1.0.5     
[...]

PR #3150, opened 2021-07-16
Java Version Validation

bpvgoncalves:
Since adoption of JEP 223, some Java versions are released only with Major version and no Minor.Security information. In such cases version is shown as X instead of X.0.0, and current regex will not match them.
Also, newer Java versions include a date as part of the version string. We must be able to differentiate numbers from dates from numbers from release, otherwise in some rare cases both version and date will be matched by the current regex resulting in failed validation due to output length.
View issue #1383 for further details.

The proposed change splits the string parsing into 2 steps.
- Trying to match date format (dddd-dd-dd) in the string and filter that out if any match exists (d = digits)
- Match within the remaining results a version in the format of 1 or 2 staring digits followed by zero or one occurrences of '.digit(s).digit(s)' pattern.

### DIFF
--- a/R/java.R
+++ b/R/java.R
@@ -94,7 +94,9 @@ validate_java_version_line <- function(master, version) {
     strsplit(versionLine, "\"")[[1]][[2]]
   } else {
     splat <- strsplit(versionLine, "\\s+", perl = TRUE)[[1]]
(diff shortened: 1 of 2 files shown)


yitao-li (approved):
Looks good to me :+1: Thanks for contributing to `sparklyr`! :slightly_smiling_face:

PR #3174, opened 2021-08-27
Use SparkEnv.get.conf to access Spark conf

joveyuan-db:


yitao-li:
Cool. Looks like we are happy with https://github.com/sparklyr/sparklyr/pull/3175 so far. I'll close this PR then.

PR #3176, opened 2021-08-31
Add support for pre-command hooks in the backend

joveyuan-db:
This PR:
* adds support for pre-command hooks in the sparklyr backend, which are intended to be run before handling any method call.  These hooks are set via an overloaded `main` method of sparklyr's `Shell` and are passed as a list of `Runnable` instances.  This will be used by Databricks to support credential passthrough (https://docs.databricks.com/security/credential-passthrough/index.html).
* adds an R unit test to verify that pre-command hooks are successfully executed.  Since there appear to be no framework for testing Scala, I added a method in `backend.scala` intended to only be used for testing.
* fixes various typos
* removes unnecessary and/or legacy code that are no longer needed to support Databricks connections


java/spark-1.5.2/shell.scala
```diff
@@ -6,6 +6,10 @@ object Shell {
   private[this] var backend: Backend = null
 
   def main(args: Array[String]): Unit = {
+    main(args, List.empty)
+  }
+
+  def main(args: Array[String], preCommandHooks: List[Runnable]): Unit = {
```
falaki, line 12: This one deserves API docs to explain `preCommandHooks` to future contributors.

java/spark-1.5.2/backend.scala
```diff
@@ -121,6 +122,10 @@ class Backend() {
     isBatch = isBatchParam
   }
 
+  def setPreCommandHooks(preCommandHooksParam: List[Runnable]): Unit = {
+    preCommandHooks = preCommandHooksParam
```
falaki, line 126:
```suggestion
    this.preCommandHooks = preCommandHooksParam
```

falaki: Also why not doing the wrapping in `Option` right here and just keep `preCommandHooksOption`

java/spark-1.5.2/backend.scala
```diff
@@ -371,7 +376,15 @@ class Backend() {
                 val backendChannel = new BackendChannel(logger, terminate, serializer, tracker)
                 backendChannel.setHostContext(hostContext)
 
-                val backendPort: Int = backendChannel.init(isRemote, port, !isWorker)
+                // create a wrapper around pre-command hooks, so that it is possible to dynamically overwrite
+                // `preCommandHooks` for testing
+                val preCommandHooksOption = Some(new Runnable {
+                    override def run(): Unit = {
```
falaki, line 382: Might be helpful to generate a log message here:
```
logger.log("Running pre-command hooks.")
```

falaki (commented):
Left some minor comments. I also asked @yitao-li to please take a look :)

yitao-li:
I'm quite excited to see this work taking place :+1:  Having credential pass thru would be very useful!

yitao-li (approved):
I agree with @falaki 's suggestions on documentation & logging, etc. Everything else looks good to me. :+1:

R/databricks_connection.R
```diff
@@ -77,11 +77,6 @@ new_databricks_connection <- function(scon, guid) {
   session <- invoke_static(sc, r_driver_local, "getDriver", guid) %>%
     invoke("sqlContext") %>%
     invoke("sparkSession")
-  # Apply user-specified fs.azure.* config values
-  fs.azure.prefix <- "fs.azure."
```
yitao-li, line 81: Truth be told, I don't think this has ever worked... Thanks for finding & deleting it!

java/spark-1.5.2/backend.scala
joveyuan-db: Discussed offline, but the `Runnable` I pass downstream needs to reference a variable local to `Backend`, so that I can dynamically modify it.  It's possible to define both `preCommandHooks` and `preCommandHooksOption` as instance variables, but I thought it would be cleaner to just construct it before I need to use it.

falaki (approved):

yitao-li:
FYI this change + other Databricks/ADL-credential-passthru-related changes after version 1.7 were released on CRAN just now as part of `sparklyr` 1.7.2.
Again thanks for contributing to `sparklyr`!

joveyuan-db:
Awesome, thanks a lot for your help @yitao-li!

PR #3182, opened 2021-09-11
dont check table existence if name randomly generated

jozefhajnala:
This PR proposes to save a call to `spark_remove_table_if_exists()` that is performed on all read operations in the updated functions if the `name` argument is `NULL`. 

If the `name` argument is not user provided (and therefore set to `NULL` by default), the [spark_read_compat_param()](https://github.com/sparklyr/sparklyr/blob/5fb0813ae06df64f8f6ef4d76c0e34d7c96206fc/R/data_interface.R#L24) function will assign a [unique, partially randomly generated](https://github.com/sparklyr/sparklyr/blob/5fb0813ae06df64f8f6ef4d76c0e34d7c96206fc/R/data_interface.R#L5) `name` to the table, meaning that we do not need to check whether the table already exists.

In my testing, this saved somewhere between 1 (for "local") and 8 (in a kubernetes-based stand-alone cluster case) seconds per function call depending on circumstances.

### DIFF
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -116,6 +116,7 @@ spark_read_csv <- function(sc,
                            memory = TRUE,
                            overwrite = TRUE,
                            ...) {
+  name_provided <- !is.null(name)
   c(name, path) %<-% spark_read_compat_param(sc, name, path)
 
   columnsHaveTypes <- length(names(columns)) > 0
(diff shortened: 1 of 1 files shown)


yitao-li (approved):
Totally makes sense! And as always, thanks for contributing to {sparklyr} :+1:

PR #3211, opened 2021-12-22
Update install_spark_windows.R

PursuitOfDataScience:
The link provided by sparklyr is expired. I've added the updated one.


yitao-li (approved):
Good catch. Thanks for the PR!

PR #3215, opened 2022-01-18
Prepare for the updated `pivot_wider()` signature

DavisVaughan:
CC @edgararuiz

We are planning to release tidyr 1.2.0 this week.

`pivot_wider()` gained 4 new arguments which you can read about [in the NEWS](https://github.com/tidyverse/tidyr/blob/main/NEWS.md#pivoting).

Unfortunately, because of an edge case in the way S3 method registration works, this causes sparklyr to break in revdep checking (in particular, this happens because of the specific combination of S3 issues outlined here https://github.com/DavisVaughan/methodtest).

There isn't anything we can do ahead of time to prevent sparklyr from breaking, but this PR is a first pass at preparing the `pivot_wider()` method for the next release. I've simply added the new arguments and then made them abort if they differ from their default value. Feel free to implement the behavior as you see fit.

Once tidyr is accepted, you should be able to bump the version requirement, remove the Remote, merge this, and then send a patch release to keep sparklyr from being removed from CRAN.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -100,7 +100,7 @@ Imports:
     rstudioapi (>= 0.10),
     stringr,
     tibble,
-    tidyr (>= 0.3.0),
+    tidyr (>= 1.1.4.9000),
     tidyselect,
     uuid,
     vctrs,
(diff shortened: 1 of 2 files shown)


edgararuiz:
Thanks @DavisVaughan !

PR #3217, opened 2022-01-21
Remove usage of deprecated function

lionel-:
`is_env()` was deprecated in rlang 0.2.0 (released in 2018-02) and it looks like the code using it wasn't covered by unit tests (otherwise you'd have received deprecation warnings). We are removing this function from rlang 1.0.0, which causes sparklyr to fail R CMD check. Sorry I didn't catch this before, it looks like sparklyr isn't included in our cloud checks.


edgararuiz:
Thanks @lionel- !!

PR #3245, opened 2022-04-06
New dpblyr compat

mgirlich:
The rendering pipeline in `dbplyr` is switching from `ops` to `lazy queries` ([Benefits](https://github.com/tidyverse/dbplyr/pull/728#issuecomment-1062612537) and [Overview of changes](https://github.com/tidyverse/dbplyr/pull/728#issue-1070334844)).
This PR makes `sparklyr` compatible with the new render pipeline.

Currently this does not yet depend on the dev version of `dbplyr`. I first wanted to be sure it still works with the CRAN version of `dbplyr`.


mgirlich:
@edgararuiz Can we work on this soonish? We would like to finally release a new dbplyr version to CRAN in the next weeks.

mgirlich:
Something weird is happening in one of the tests (test-ml-regression-linear-regression.R -> weights column works for lm).
I got a warning about a partial argument match (`partial argument match of 'features' to 'features_col'`). So, I changed `features` to `features_col`. But this produced an error.
This is not connected to this PR but probably you should have a look at that.

edgararuiz:
@mgirlich - Hi, I see several calls for unexported `dbplyr` functions, such as `add_select()` and `add_summarise()`. I don't think I'll be able to publish to CRAN with `dbplyr:::add_select()`.  Any ideas on how to handle?

edgararuiz:
It looks like there are a couple of failures when testing with `dbplyr` 1.4.4: 

```
══ Failed ═════════
── 1. Failure (test-dplyr.R:557:3): summarise(.groups=) ───────────────────────────────────
sdf %>% summarise(.groups = "drop") %>% group_vars() not equal to character().
Lengths differ: 1 is not 0

── 2. Failure (test-dplyr.R:558:3): summarise(.groups=) ───────────────────────────────────
sdf %>% summarise(.groups = "keep") %>% group_vars() not equal to c("x", "y").
Lengths differ: 1 is not 2
```

You can see them by running: 

```
devtools::install_version("dbplyr", version = "1.4.4")
devtools::test(filter = "^dplyr$", reporter = sparklyr_reporter())
```

I'll work on adding a test run in GHA for `dbplyr` version  < 2, to get this check automated.

edgararuiz:
It looks like those tests also fail on a run against the `main` branch, so no need to address them(https://github.com/sparklyr/sparklyr/runs/6118405455?check_suite_focus=true#step:14:1650)

mgirlich:
With some small changes to dev dbplyr (https://github.com/tidyverse/dbplyr/pull/834) sparklyr can use much more of the dbplyr code instead of copy-pasting the code. This should also provide some new features for free to sparklyr.
This should now hopefully work well for the CRAN and the dev version of dbplyr.

tests/testthat/test-dplyr.R
```diff
@@ -535,6 +537,8 @@ test_that("result from dplyr::compute() has remote name", {
 })
 
 test_that("dplyr::summarize() emits an error for summarizer using one-sided formula", {
+  skip_if_not(dbplyr_uses_ops())
```
mgirlich, line 540: This works now because it uses dbplyr. Is there any good reason why this should not work?

tests/testthat/test-dplyr.R
```diff
@@ -131,6 +131,8 @@ test_that("'across()' works with formula syntax", {
 })
 
 test_that("'across()' emits error when formula references a mutated column", {
+  skip_if_not(dbplyr_uses_ops())
```
mgirlich, line 134: This works now because it uses dbplyr. Is there any good reason why this should not work?

R/dplyr_spark_connection.R
```diff
@@ -17,22 +17,43 @@ fix_na_real_values <- function(dots) {
...
-    partial_eval_dots(sim_data = simulate_vars(.data))
+  if (dbplyr_uses_ops()) {
+    dots <- rlang::enquos(..., .named = TRUE) %>%
+      fix_na_real_values() %>%
+      partial_eval_dots(sim_data = simulate_vars_spark(.data))
 
-  nest_vars(.data, dots, character())
+    nest_vars(.data, dots, character())
+  } else {
+    dots_org <- rlang::enquos(..., .named = TRUE)
```
mgirlich, line 29: Unfortunately, this hack is required to use the `dbplyr` method.

edgararuiz: Not sure if you need me to do anything here, I'm ok with this change

edgararuiz:
The failing tests on for the 1.4 `dbplyr` are unrelated to this change, so I'm ok with merging  this PR if you are

R/dplyr_spark_connection.R
mgirlich: Good, just wanted to let you know.

mgirlich:
I added a dependency on the dev version of dbplyr to see whether the tests still work (they did locally but I only tested one setup). Can you run the tests again?

edgararuiz:
> @edgararuiz I fixed an issue in dev dbplyr. Can you run the tests again?

Hi, I don't see a new commit, so I can't start the checks. Did I miss anything?

mgirlich:
No, you didn't miss anything. I needed to fix something in dbplyr, not in sparklyr. I quickly added a NEWS entry so that you can run the tests.

edgararuiz:
@mgirlich - To be honest, I wouldn't feel comfortable sending `sparklyr` to CRAN with a dev dependency on `dbplyr`, is there an alternative?

mgirlich:
Sorry for the misunderstanding. I don't want to send it to CRAN with the dev dependency. I just wanted to see the test results. When the issues are fixed I will remove the dependency on dev dbplyr.

There are now 3 different failures left:
* [Spark 2.2](https://github.com/sparklyr/sparklyr/runs/6199357211?check_suite_focus=true)

```
extraneous input '<' expecting {'SELECT' ...
...
== SQL ==
MAP<STRING, INT>
---^^^
```

This looks more like an issue with spark or sparklyr to me

* [Spark 3.1 with Arrow Dev](https://github.com/sparklyr/sparklyr/runs/6199356942?check_suite_focus=true): I could not reproduce the issue locally (though I did not manage to install Arrow Dev) 
* [Spark 3.1 with dbplyr 1.4](https://github.com/sparklyr/sparklyr/runs/6199356802?check_suite_focus=true): Error : object 'op_grps' not found whilst loading namespace 'sparklyr': Is it still necessary that sparklyr works with dbplyr 1.4? dbplyr 2.0 was released ~ 18 months ago and has some nice improvements.

@edgararuiz Can you help with these issues? After they are fixed the dependency on dev dbplyr can be removed and I think this can be merged.

edgararuiz:
@mgirlich - Hey, I just committed a change to main, which removes the testing on the old `dbplyr` version.  I think we are ok with dropping support for `dbplyr` < 2 in `sparklyr`.  This test run will also help us determine if the failure in Arrow Dev is because changes on the PR vs changes in the Arrow source code

mgirlich:
@edgararuiz I removed the remote dependency. So, I don't see anything left to be done.
I also changed `select()` to use the dbplyr method if the new dbplyr is installed.

mgirlich:
@edgararuiz We added `simulate_vars_is_typed()`. This will allow (after a tidyselect release) to disallow predicate selections for lazy tables but still allow them for spark tables. But now this should be finished 😄

edgararuiz:
Thank you @mgirlich for all the work on this!

PR #3263, opened 2022-05-20
Simplifies ML functions

edgararuiz:
- Centralizes model processing into a few functions
- Switches Regressor to new model processing


t-kalinowski (approved):
LGTM!

This significantly reduces duplication of arguments, so this is great.

I had a few minor tweaks and inline comments. I pushed them to a fork, you can merge it into this branch if you'd like: https://github.com/t-kalinowski/sparklyr/tree/simplifyml

One thought I had while looking at this: does it make sense to get rid of the S3 methods for `spark_connection`, `ml_pipeline`, and `tbl_spark`, since everything goes through the default method? If yes, then maybe adding a
```r
stopifnot(c("spark_connection", "ml_pipeline", "tbl_spark") %in% class(x))
``` 
might be a good guard to add to the default method.

GJ!

PR #3264, opened 2022-05-23
Compatibility with dev dbplyr

mgirlich:
Some more small changes are required for the compatibility with dev dbplyr.

@edgararuiz 
We're planning to release dbplyr to CRAN on June 3. A CRAN release of sparklyr until then would make sense. Can you manage to release soon?


mgirlich:
@edgararuiz The test for `nest()` produces a different variable order with dev dbplyr. It is the same as for data frames so this seems to be an improvement 😄  Now, the tests should pass again...

edgararuiz:
Thanks @mgirlich !

PR #3272, opened 2022-06-24
Add spark_insert_table function

jimhester:
This allows one to insert data into an existing table definition without
redefining the table, even when overwriting the existing data.

The first test illustrates the problem with the existing method. The existing table definition is ignored by `spark_write_table(mode = "overwrite")`. This causes problems if you are trying to follow best practices of having a table definition defined upfront before writing to it.

Incidentally 88318ab6dfba4ba6e0f0140ec1ada4b45e969e7c should also fix at least one instance of https://github.com/sparklyr/sparklyr/issues/3213, which is caused not by bigints, but because the code currently calls `as.integer(Inf)`, which produces the warning. That commit is tangential to my main purpose of this PR, but thought it worth fixing here anyway.


jimhester:
cc @edgararuiz, as I am not sure you are currently watching this repo on GH, so may miss this. Could you also enable the CI workflow for this PR?

edgararuiz:
Thank you!

PR #3277, opened 2022-07-12
Removes `dbplyr` backwards compatibility code

edgararuiz:
Since 1.76, `sparklyr` took a hard dependency on `dbplyr` version 2.1.1 or above. This means that backwards compatibility with anything that supports ops is no longer necessary. 

This PR removes the code used to provide support for older versions of `dbplyr` while the transition occurred.  It also add a number of tests to provide better coverage for `dplyr` and `dbplyr` related operations.

By bumping `dbplyr` dependency a big more, to 2.2.1, we are able to remove all of the code switching between "uses ops" and not.


NEWS.md
```diff
@@ -4,6 +4,9 @@
 
 - Adds support to Spark 3.3 
 
+- Removes code related to backwards compatibility with `dbplyr`. `sparklyr`
+  requires `dbplyr` version 2.1.1 or above, so the code is no longer needed.
```
mgirlich, line 8:
```suggestion
requires `dbplyr` version 2.2.1 or above, so the code is no longer needed. 
```

PR #3283, opened 2022-08-05
Remove workaround for arrow < 0.17

nealrichardson:
`arrow` 9.0.0, landing on CRAN next week (one hopes), will remove the long-deprecated `arrow::write_arrow()` function. We upgraded `sparklyr` not to use that long ago, but it is still referenced in the code here for compatibility. Unfortunately, `R CMD check` will now throw a warning on sparklyr: `Missing or unexported object: 'arrow::write_arrow'`, even though the code itself is never exercised. 

I've updated the R code but did not run the `update_embedded_sources.R` script that the comment mentions, as the machine I'm on doesn't seem to have the right Java setup. Can you easily do that and push to my PR?


edgararuiz:
I'm just curios about `arrow` 9.0.0. It looks like `write_arrow()` is still exported by the package. That is why my other commits were not failing after those checks started using the most recent CRAN version of `arrow`.  Is the plan to remove it on a later version?

edgararuiz:
Also, I'm ready to merge if you are. I need to send an update to CRAN, tomorrow if possible. And I'll be happy to include this change if you like.

nealrichardson:
I added back a dummy write_arrow when I submitted so that our revdep checks didn't fail, and/or so you wouldn't get an urgent nasty email about fixing it. So from our side, this just needs to reach CRAN before we release again in 3 months.

nealrichardson:
Feel free to merge when you're satisfied with this. Thanks for the help!

edgararuiz:
Thank you!

akshaykjain:
does [sparklyr-master-2.12.jar](https://github.com/sparklyr/sparklyr/blob/main/inst/java/sparklyr-master-2.12.jar) support Spark 3.3.0 now?

PR #3295, opened 2022-10-10
Adding support for 3 tier namespaces `<catalog>.<schema>.<table>`

zacdav-db:
Adjusting the `process_tbl_name()` function to support 3 tier namespaces.

Small change, works on Databricks when built and tested - this is required for those who wish to use 'Unity Catalog' which has a 3 tier namespace.

Currently there are no tests specific to `spark_read_table()` only `spark_write_table()`.
Happy to take guidance on tests that should be added (if any).

Signed-off-by: Zac Davies <zac@databricks.com>

### DIFF
--- a/R/dplyr_spark.R
+++ b/R/dplyr_spark.R
@@ -138,8 +138,10 @@ process_tbl_name <- function(x) {
         x
       } else if (identical(num_components, 2L)) {
         dbplyr::in_schema(components[[1]], components[[2]])
+      } else if (identical(num_components, 3L)) {
+        dbplyr::in_catalog(components[[1]], components[[2]], components[[3]])
       } else {
-        stop("expected input to be <table name> or <schema name>.<table name>")
+        stop("expected input to be <table name>, <schema name>.<table name>, or <catalog name>.<schema name>.<table name>")
       }
     }
   }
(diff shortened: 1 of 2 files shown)


falaki (commented):
Thanks for submitting this change. Would you look into adding a unit test for it?

falaki (approved):

zacdav-db:
Test are failing trying to access cloudfront.

```
Error in `download.file(...)`: cannot open URL 'https://d3kbcqa49mib13.cloudfront.net/spark-1.6.3-bin-hadoop2.6.tgz'
```

zacdav-db:
The URL's in https://github.com/sparklyr/sparklyr/blob/main/inst/extdata/versions.json are seemingly the culprit, specifically those versions that are using the cloudfront domain.

@javierluraschi I believe you created the file initially - any advice or additional context you can share would be appreciated.

falaki:
@edgararuiz is this something you can fix?

edgararuiz:
Yeah, I just pushed an update to `main` that updates the JSON file. I also made an update in the script that keeps track of where to download which version of Spark, to reflect a new prioritization.  I'd suggest to merge in this change once it passes GHA checks.

PR #3335, opened 2023-05-21
Adding support of parameterized queries from Spark 3.4.0

gregleleu:
Spark 3.4.0 added parameterized queries natively (vs currently interpolating the SQL).
This PR adds support for that with an added `params` argument for dbSendQuery (passed to dbGetQuery).
The params argument should be in line with what other interfaces are doing: https://dbi.r-dbi.org/articles/dbi-advanced#how-to-use-parameters-safely-in-sql-queries

Also: is anything else interesting to add from 3.4.0? Spark Connect?


tests/testthat/test-dbi.R
```diff
@@ -36,6 +36,15 @@ test_that("dbGetQuery works with parameterized queries", {
   expect_equal(nrow(virginica), 50)
 })
 
+test_that("dbGetQuery works with native parameterized queries", {
+  if (spark_version(sc) < "3.4.0") skip("native parameterized queries not supported before 3.4.0")
```
edgararuiz, line 40: Can we use `test_requires_version("3.4.0")` instead? I can commit this change if you're ok with that.

edgararuiz:
Thank you!

Issue #3331, opened 2023-04-19
Cannot install spark 3.2.3


mhamilton723:
`spark_install(version = "3.2.3", hadoop_version = "3.2")` 

fails on sparklyr 1.8.1 with the following error

```Error in download.file(...) : 
  cannot open URL 'https://dlcdn.apache.org/spark/spark-3.2.3/spark-3.2.3-bin-hadoop3.2.tgz'
```

To fix this sparklyr needs to pull from this URL instead

`https://archive.apache.org/dist/spark/spark-3.2.3/spark-3.2.3-bin-hadoop3.2.tgz`

edgararuiz:
Fixed c32284916e8f36600ed2f70c4569a523163e91d1

PR #3340, opened 2023-06-20
Updates

edgararuiz:
- Adds Spark 3.4 to downloads 
- Updates jars (may back this change out if tests fail)
- Fixes tibble print tests


edgararuiz:
Fixes #3331

PR #3344, opened 2023-06-22
Compatibility with dev dbplyr

mgirlich:


edgararuiz:
@mgirlich - Hi, I'd like to do a release to CRAN soon. Should I hold off until `dbplyr` is released, in case there are more changes needed?

mgirlich:
As the CRAN sparklyr version isn't compatible with the dev version of dbplyr, it would be great if you could release before dbplyr. I'll let you know until next week whether I see other possibly breaking changes coming up.

PR #3346, opened 2023-06-23
Fix early return on batch jobs

### LINKED ISSUE #3311: `spark_submit(...)` not terminated after submission
[This line](https://github.com/sparklyr/sparklyr/blob/eb3e795447887908d9e795512ad08eeeb32eede5/R/shell_connection.R#L442) is supposed to terminate batch connections before going any further. However, `batch` is the file name from `spark_submit()`. Not sure if this should be switched to `isBatch` instead, or I am missing something.

smingerson:
Early return on `isBatch`, which is the logical value, instead of `batch`, while is the file name. Closes #3311.

### DIFF
--- a/R/shell_connection.R
+++ b/R/shell_connection.R
@@ -440,7 +440,7 @@ start_shell <- function(master,
   }
 
   # batch connections only use the shell to submit an application, not to connect.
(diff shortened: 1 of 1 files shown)


edgararuiz:
Thank you!

PR #3419, opened 2024-02-07
Update data_interface.R

ahcyip:
I added notes about how to specify columns and infer_schema, but this definitely needs review.

This is how I believe it works based on my review of the code, but the project/repo maintainer should definitely verify/correct where necessary.

### DIFF
--- a/R/data_interface.R
+++ b/R/data_interface.R
@@ -62,14 +62,17 @@ spark_csv_options <- function(header,
 #'   is, should the table be cached?)
 #' @param header Boolean; should the first row of data be used as a header?
 #'   Defaults to \code{TRUE}.
-#' @param columns A vector of column names or a named vector of column types.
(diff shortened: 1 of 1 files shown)


ahcyip:
OK, reviewing again, I'm not sure anymore. So I deleted the uncertain parts, but the writer/maintainer of spark_read_x should confirm/clarify these 2 points here:

#' @param columns
#'   Columns without type specification will be inferred if infer_schema = TRUE or excluded if infer_schema = FALSE.
#' @param infer_schema
#'   Column types that cannot be automatically inferred will default to StringType.

ahcyip:
OK... actually, a partial list will rewrite the names of the first columns up to the length of the partial list and thereby likely refer to the wrong columns. So it seems like a partial columns list would not be useful and could be misleading / produce unexpected results for most users. Perhaps sparklyr should disable this partial list option completely.

PR #3432, opened 2024-04-09
Streaming via CloudFiles and Tables

zacdav-db:
Adding support for

- Databricks "[autoloader](https://docs.databricks.com/en/ingestion/auto-loader/options.html)" (format: `cloudFiles`) for streaming ingestion of files(`stream_read_cloudfiles`)
- `stream_write_table`
- `stream_read_table`

Made changes to `stream_write_generic`:

- `toTable` method doesn't allow calling `start`, added `to_table` param that adjusts logic
- `path` option not propagated when `to_table` is `TRUE`


edgararuiz:
Also, would it be able to re-document with `roxygen2` version 7.3.0, that way the PR will be more focused to the stream functions additions.  I can take care of the `roxygen2` upgrade in a different PR. Thanks!

edgararuiz:
Thanks!

PR #3454, opened 2024-08-15
Fix warnings with dbplyr and raw columns

gregleleu:
Code was trying to do as.raw(NA) which creates a warning.
Replaced with a version that does creates a zero-row tibble with the right types.

Should fix https://github.com/sparklyr/sparklyr/issues/3357


edgararuiz:
Thanks!

PR #3460, opened 2024-09-05
fix: `src_databases`,column with name 'databaseName' cannot be resolved

weichisyu:
Try adding a new argument to handle situations where the column name is not databaseName.


edgararuiz:
Thanks!

PR #3470, opened 2024-10-28
Allow setting path to NA in spark_read_source()

### LINKED ISSUE #3469: Cannot call spark_read_source with a NULL path
I would like to call `spark_read_source` with a `NULL` path, i.e. `spark_read_source(sc, name, path = NULL, ...)`. Looking at the implementation of `spark_read_source`, this seems to be an intended use case:

```r
...
df <- if (is.null(path)) {
  invoke(df_reader, "load")
}
...
```

The intention behind using a `NULL` path is basically to invoke `load()` on the underlying data frame reader instead of `load(...)` with the path.

The problem, however, is that a `NULL` path does not seem to work with `spark_read_compat_param` as this will lead to the following error: `Error in `c(name, path) %<-% spark_read_compat_param(sc, name, path)`: Assignment must contain same number of elements`

Example
Instead of 
```r
spark_read_source(sc, name="test", path = "test.csv", source="csv")
```
I would like to write
```r
spark_read_source(sc, name="test", path = NULL, source="csv", options=list(path = "test.csv"))
```

The latter can directly be used as a reproducible example since even though the files may not exist, this will immediately fail with
```
Error in c(name, path) %<-% spark_read_compat_param(sc, name, path) : 
  Assignment must contain same number of elements
```

martinstuder:
Allow setting path to `NA` in `spark_read_source()`. Fixes #3469


martinstuder:
@edgararuiz Let me know if this looks sensible to you or if you would prefer fixing the issue using a different approach.

martinstuder:
Would be great to have a solution for #3469. Happy to incorporate any feedback.

PR #3502, opened 2025-09-15
ml_load(): use Spark SQL glob for metadata (supports UC /Volumes); keep legacy textFile fallback

AMRUTH-ASHOK:
Problem
`ml_load()` fails to load models from Unity Catalog Volume paths (/Volumes/...) with IllegalArgumentException.

Root cause
`SparkContext.textFile(..., "metadata/part-00*")` path can be treated like local FS and trip UC access checks.
Spark SQL / DataFrameReader is UC-aware and resolves /Volumes correctly.

Change
- Prefer Spark JSON glob: `file.path(path, "metadata", "part-*")`, parse `class`, then `class.load(path)`.
- Keep the original textFile path as a fallback to avoid breaking edge cases.

Tests
- `tests/testthat/test-ml-load.R`: happy-path (glob) + forced-failure fallback.

No API changes; behavior is preserved on non-UC paths.


R/ml_persistence.R
```diff
@@ -85,15 +85,45 @@ ml_save.ml_model <- function(x, path, overwrite = FALSE,
...
+      name = metadata_view,
+      path = metadata_glob
+    ),
+    silent = TRUE
+  )
+
+  if (!inherits(df_try, "try-error")) {
+    class_val <- NULL
+
+    cls1 <- try(dplyr::pull(df_try, "class"), silent = TRUE)
```
edgararuiz, line 109: I think if we can change this `try()` call with  `df_try_cols <- colnames(df_try)` and then interrogate here, and on line 114 simply check if `df_try_cols` contains those columns. Is there a specific reason to use `try()`?

R/ml_persistence.R
```diff
@@ -102,11 +132,19 @@ ml_load <- function(sc, path) {
...
 
-  if (!inherits(metadata, "try-error")) {
-    class <- metadata$class
-    invoke_static(sc, class, "load", path) %>%
-      ml_call_constructor()
-  } else {
-    stop("ML could not be loaded:\n  ", metadata)
+  if (!inherits(metadata_try, "try-error")) {
+    class_val <- metadata_try$class
+    if (is.null(class_val) || length(class_val) == 0 || is.na(class_val) || !nzchar(class_val)) {
```
edgararuiz, line 137: When would `class_val` be `NA`?

edgararuiz (commented):
Thank you @AMRUTH-ASHOK ! I just left a couple of comments. Also, don't worry about the GHA failures, they are not due to something you'd have to update.

edgararuiz:
Thanks!

